PR #5, opened 2016-08-24
Fixed javadoc errors.

yeunga:


opencadc-admin:
Then gradle assemble and check tasks are redundant since build does them. Just "build javadoc install" should suffice.

yeunga:
Fixed. Thanks.
________________________________________

PR #15, opened 2016-11-10
Fix to support TERENA certificates

bertocco:
Fix to support TERENA certificates. Already present in cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/UserServlet.java

### DIFF
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/ResetPasswordServlet.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/ResetPasswordServlet.java
@@ -104,6 +104,8 @@ import ca.nrc.cadc.auth.ServletPrincipalExtractor;
 import ca.nrc.cadc.log.ServletLogInfo;
 import ca.nrc.cadc.net.TransientException;
 import ca.nrc.cadc.util.StringUtil;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Servlet to handle password resets.  Passwords are an integral part of the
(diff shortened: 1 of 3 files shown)


pdowler:
Sara,

I understand the changes except for the regex that parses the string from the web.xml .... can you add some text to the class javadoc for the two effected servlets to explain how to configure the servlets now? I hope it is a  simple as multi-line config with line breaks between the trusted DNs.

Note that the travis CI also builds javadocs (because they are required for publishing to maven) and documenting xml in there is notoriously tricky. Make sure to check that the javadocs build without errors and also check the results of the build in the pull request (after you modify) because we have Travis CI running with 2 different java versions (and they differ in javadoc pickiness).

bertocco:
The patch is basically cadc based. 
The same piece of code originally was present in 3 servlets: 
ca/nrc/cadc/ac/server/web/UserServlet.java
ca/nrc/cadc/ac/server/web/UserRequestServlet.java
ca/nrc/cadc/ac/server/web/ResetPasswordServlet.java
during our tests the use of terena certificates revealed one problem in certificate subject parsing, so someone at cadc patched UserServlet.java but the problem is still present in UserRequestServlet.java and ResetPasswordServlet.java 
I simply cut and pasted the patch present in UserServlet.java  to UserRequestServlet.java and ResetPasswordServlet.java

In any case I'll write the javadoc for the 3 servlets and I'll pull it.

pdowler:
Sara,

Do you get a notification when the travis CI fails? Anyway, it failed for this pull and as I expected the java 8 javadoc is much pickier than 7. If you click on "Details" in the failed build you can go to the travis page and click on the failed build to get the console output... or install jdk 8 on your system so you can build and diagnose locally.

PR #28, opened 2017-02-01
Fixes #27

### LINKED ISSUE #27: GMSClient always uses the CERT AuthMethod.
The `ca.nrc.cadc.ac.client.GMSClient` class in `cadc-access-control` always default to the CERT AuthMethod in each of its calls.  This needs to be modified to allow Cookie access from the UI.

at88mph:


cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -1179,4 +1125,21 @@ protected RegistryClient getRegistryClient()
...
+            {
+                if (o instanceof X509CertificateChain)
+                    return AuthMethod.CERT;
+                if (o instanceof SSOCookieCredential)
+                    return AuthMethod.COOKIE;
+                // AuthMethod.PASSWORD not supported
+                // AuthMethod.TOKEN not supported
+            }
+        }
+        return AuthMethod.ANON;
```
brianmajor: throw an AccessControlException here instead of returning ANON since there is no anonymous access to the group capabilities.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -822,10 +768,10 @@ private Principal getCurrentUserID()
         StringBuilder searchGroupPath = new StringBuilder("?");
         //searchGroupURL.append("ID=").append(NetUtil.encode(id));
         //searchGroupURL.append("&IDTYPE=").append(NetUtil.encode(idType));
-        searchGroupPath.append("&ROLE=").append(NetUtil.encode(roleString));
+        searchGroupPath.append("ROLE=").append(NetUtil.encode(roleString));
```
brianmajor: Hmm, I guess this just broken before when the ID & IDType were removed?

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -89,6 +89,8 @@
 import javax.net.ssl.SSLSocketFactory;
 import javax.security.auth.Subject;
 
+import ca.nrc.cadc.auth.*;
+import ca.nrc.cadc.net.*;
```
brianmajor: I'm guessing your IDE changes the imports to * when there are more than 4 references :)  I prefer to see the full list mostly for consistency in this project.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -431,18 +423,21 @@ public Group updateGroup(Group group)
...
-            return groupReader.read(retXML);
-        }
-        catch (Exception bug)
-        {
-            log.error("Unexpected exception", bug);
-            throw new RuntimeException(bug);
-        }
+        return getGroup(group.getID().getName());
+
+// Cookie gets lost when following redirect and pulling the XML down!
```
brianmajor: If the cookie gets lost, isn't that a bug in HttpPost that should be fixed?

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -457,55 +452,37 @@ public void deleteGroup(String groupName)
...
-        }
+        HttpDelete delete = new HttpDelete(deleteGroupURL, true);
+        delete.setSSLSocketFactory(getSSLSocketFactory());
+        delete.run();
 
-        if (responseCode != 200)
+        Throwable error = delete.getThrowable();
+        if (error != null)
         {
-            String errMessage = NetUtil.getErrorBody(conn);
```
brianmajor: I'm not sure if the text in getErrorBody() and throwable.getMessage() are the same.  If different, then the error reported on failed deletes will change.  Probably no biggie unless there's an integration test expecting a certain response.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -1179,4 +1092,54 @@ protected RegistryClient getRegistryClient()
...
+                if (o instanceof SSOCookieCredential)
+                    return AuthMethod.COOKIE;
+                // AuthMethod.PASSWORD not supported
+                // AuthMethod.TOKEN not supported
+            }
+
+            throw new AccessControlException("No valid public credentials.");
+        }
+        else
+        {
```
brianmajor: Since the GMSClient has knowledge of the capabilities of the service, it knows that it does not support anonymous access, so I would throw an AccessControlException here too (instead of returning ANON).

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -1179,4 +1092,54 @@ protected RegistryClient getRegistryClient()
...
+     *                                      provided AuthMethod.
+     */
+    private URL lookupServiceURL(final URI standard)
+            throws AccessControlException
+    {
+        final URL serviceURL = getRegistryClient()
+                .getServiceURL(this.serviceID, standard, getAuthMethod());
+
+        if (serviceURL == null)
+        {
```
brianmajor: (this goes with the comment below on getAuthMethod())  -  I'd throw a general RuntimeException here because the service doesn't have a capability the client is expecting....

at88mph:
I swear if this changes the tests... :)

at88mph:
Changes complete.

PR #44, opened 2018-02-20
use caller subject as cache for NumericPrincipal to Subject 

pdowler:
Note: file reformated to current standards - relevant change is the if block at line 225


brianmajor:
The unit tests fail because of the new check for the header X-Forwarded-For in cadc-log.  Mocking unit tests around HTTPServerRequest was and the request handling was probably not a good idea.  For now, to allow extra calls to the request object, we can switch the tests to use EasyMock.createNiceMock() instead of EasyMock.createMock().

pdowler:
sigh. I hate EasyMock

PR #47, opened 2018-04-11
fixed scalability bug with getAugmentedUser 

pdowler:
only get memberOf attr when callers says to prime GroupMembership cache


cadc-access-control-server/src/test/java/ca/nrc/cadc/ac/server/ldap/AbstractLdapDAOTest.java
```diff
@@ -132,107 +132,113 @@ public static void setUpBeforeClass()
         throws Exception
     {
         Log4jInit.setLevel("ca.nrc.cadc.ac", Level.INFO);
+        Log4jInit.setLevel("ca.nrc.cadc.util", Level.DEBUG);
```
brianmajor: should be set back to INFO

PR #48, opened 2018-04-26
s2247: adapt to Delegation Token & related changes in cadc-util 1.1.2

hjeeves:


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/LoginServlet.java
```diff
@@ -185,6 +186,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+                // Get set of all principals that apply to the user.
+                // Cookie will have all principals added to it.
+                AuthenticatorImpl ai = new AuthenticatorImpl();
+                Subject userSubject = new Subject();
+                userSubject.getPrincipals().add(p);
+                ai.augmentSubject(userSubject);
+                Set<Principal> userPrincipals = userSubject.getPrincipals();
+
+                // domains have to be harvested here somehow.
+                // what is the default domain
```
brianmajor: This comment probably doesn't apply any more?  That work is done in the cookie manager.

cadc-access-control/build.gradle
```diff
@@ -24,7 +24,7 @@ dependencies {
     compile 'org.jdom:jdom2:2.+'
     compile 'org.json:json:20160212'
 
-    compile 'org.opencadc:cadc-util:1.+'
+    compile 'org.opencadc:cadc-util:[1.1,)'
     compile 'org.opencadc:cadc-cdp:[1.0,)'
     compile 'org.opencadc:cadc-registry:1.+'
```
brianmajor: I don't think you need to update the build file here.  No changes so no version update.  It will still get the latest cadc-util.

cadc-access-control-identity/src/main/java/ca/nrc/cadc/auth/ACIdentityManager.java
```diff
@@ -247,6 +248,20 @@ public Subject toSubject(Object o) {
...
     public void augmentSubject(final Subject subject) {
+
+        if (subject == null ) {
+            return;
+        }
+
+        // If the principal list is in the subject has  Numeric Principal
+        // AND the list is greater than 1, then LDAP doesn't need to be
+        // called here
+        Set<Principal> principalSet = subject.getPrincipals();
```
brianmajor: This line is no longer used so should be deleted

cadc-access-control-identity/src/main/java/ca/nrc/cadc/auth/ACIdentityManager.java
```diff
@@ -247,6 +248,20 @@ public Subject toSubject(Object o) {
...
+        Set<NumericPrincipal> nPrincipalSet = subject.getPrincipals(NumericPrincipal.class);
+        if (principalSet != null && principalSet.size() > 1 && nPrincipalSet != null ) {
```
brianmajor: The API to subject.getPrincipals says that it creates a new Set each invocation, so this line can be as simple as if (!principalSet.isEmpty())

cadc-access-control/build.gradle
```diff
@@ -15,7 +15,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.11'
+version = '1.1.12'
```
brianmajor: Why a new version of this library?

PR #51, opened 2018-05-22
s2255: getUser and getAugmented user function with multiple ldap entries returned.

hjeeves:


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -525,61 +610,82 @@ private User getUser(final Principal userID, final String usersDN)
...
-            user.personalDetails.city = searchResult.getAttributeValue(LDAP_CITY);
-            user.personalDetails.country = searchResult.getAttributeValue(LDAP_COUNTRY);
-            user.personalDetails.email = searchResult.getAttributeValue(LDAP_EMAIL);
-            user.personalDetails.institute = searchResult.getAttributeValue(LDAP_INSTITUTE);
+            LdapDAO.checkLdapResult(e.getResultCode());
         }
 
-        logger.debug("getUser: found " + userID.getName() + " in " + usersDN);
-        return user;
+        logger.debug("getUserList returning " + userList.size() + " entries for " + userID.toString());
```
andamian: Is this the only place where multiple entries are logged? If so, shouldn't it be a higher level of debugging?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -351,6 +353,14 @@ private String getEmailAddress(final User user)
     private void checkUsers(final Principal userID, final String email, final String usersDN)
         throws TransientException, UserAlreadyExistsException
     {
+        // Internal system property set during LdapUserDAOTest.testAddMultipleUser
+        // so that this check can be overridden
+        if (System.getProperty(SUPPRESS_CHECKUSER_KEY) != null && System.getProperty(SUPPRESS_CHECKUSER_KEY).equals("true"))
+        {
```
andamian: Not a very big fan of this approach. Subclassing LdapUserDAO in the test code and overriding this method would have achieved the same effect without creating this backdoor.

cadc-access-control-server/src/test/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAOTest.java
```diff
@@ -849,7 +852,155 @@ public Object run() throws Exception
...
+                    // Check to see if there are multiple users already existing.
+                    List<User> userList = userDAO.getAllUsers(userID, config.getUsersDN());
+
+                    if (userList.size() < 6) {
+                        log.debug("Creating multiple users (5 more)");
+
+                        // Run this 5x to create multiple users
+                        for (int i = 0; i < 5; i++) {
+                            try {
+                                userDAO.addUser(testUser);
```
andamian: It appears to me that the testUser are all identical (with or without HttpPrincipal) which is not exactly the main scenario of one real user with multiple duplicates that don't have HttpPrincipal

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
hjeeves: I hear you. This was at Pat's suggestion. It was something that isn't supposed to be documented or available publicly. It's used only in testing.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -485,37 +495,112 @@ public User getUserRequest(final Principal userID)
...
+
+            // 'Real' user will have an LDAP_USER_NAME that is NOT EXTERNAL_USER_CN
+            if (EXTERNAL_USER_CN.equals(username)) {
+                continue;
+            }
+            ret = next;
+            break;
+        }
+
+        if (ret == null) {
```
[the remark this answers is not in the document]
hjeeves: Good point. Thanks. :)

cadc-access-control-server/src/test/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAOTest.java
hjeeves: The key thing is userDAO.addUser is the internal method (inserts user directly into final tree, with $EXTERNAL-CN as LDAP-USER-NAME. the 'addUserRequest/approveUserRequest' cycle inserts the user information correctly into the ldap entry. HttpPrincipal existing or not isn't part of the check.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -525,61 +610,82 @@ private User getUser(final Principal userID, final String usersDN)
...
                 logger.debug(msg);
                 throw new UserNotFoundException(msg);
             }
-        }
-        catch (LDAPException e)
+
+            // Determine which is the 'real' user (not automatically generated
+            // by vospace interaction, for example,) and return that.
+            foundUser =  getUserFromResultList(multiSearchResult);
+            logger.debug("getUser: found " + userID.getName() + " in " + usersDN);
```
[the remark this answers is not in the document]
hjeeves: okay

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -485,37 +495,112 @@ public User getUserRequest(final Principal userID)
...
+
+        if (StringUtil.hasLength(firstName) && StringUtil.hasLength(lastName)) {
+            newUser.personalDetails = new PersonalDetails(firstName, lastName);
+            newUser.personalDetails.address = userEntry.getAttributeValue(LDAP_ADDRESS);
+            newUser.personalDetails.city = userEntry.getAttributeValue(LDAP_CITY);
+            newUser.personalDetails.country = userEntry.getAttributeValue(LDAP_COUNTRY);
+            newUser.personalDetails.email = userEntry.getAttributeValue(LDAP_EMAIL);
+            newUser.personalDetails.institute = userEntry.getAttributeValue(LDAP_INSTITUTE);
+        }
+
```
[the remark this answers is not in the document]
hjeeves: okay

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
hjeeves: Another good point. This particular function is part of the test code (and might get percolated to the cadc-access-control-admin command line interface at a later step.) It's not the only one that will detect multiple ldap entries. There isn't an explicit message blurted if more than one is found (unless it's flagged as an error.) - I think that's a ball drop on my part.  For the sake of tracking the error, I think an info level would be most likely to be seen. ?

cadc-access-control-server/src/test/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAOTest.java
andamian: OK, I see. But I would say it's a bit risky to have the method you are testing (getUser) part of the setup of the test (line 889).

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
andamian: Not documented but available publicly. Why is the initial value "cadc.skip.checkuser" and not just null? Does it have a meaning?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserDAO.java
```diff
@@ -507,6 +507,10 @@ public User getUserRequest(final Principal userID)
     private SearchResultEntry getFirstUserEntry(SearchResult multiSearchResult) {
         SearchResultEntry ret = null;
 
+        if (multiSearchResult == null) {
```
andamian: how about size()=0?

PR #54, opened 2018-09-07
S2325

yeunga:
Reverted previous changes since the unit test failures were the results of a bug in JsonInputter as pointed out by Pat.

Removed usage of JsonInputter.getListElementMap() since this was no longer available. This was a way to recover the name of an item in an array of items. JsonInputter can now recover the item name without the need to use the listElementMap.


yeunga:
Depends on pull request https://github.com/opencadc/core/pull/63

PR #59, opened 2019-01-31
S2243

yeunga:
Updated cadc-access-control and cadc-access-control-server modules to allow for multiple readOnly LDAP servers.


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/GroupPersistence.java
```diff
@@ -146,12 +146,14 @@ void deleteGroup(String groupID)
...
      *
+     * @return A Group instance
+     *
      * @throws GroupNotFoundException If the group was not found.
      * @throws TransientException If an temporary, unexpected problem occurred.
      * @throws AccessControlException If the operation is not permitted.
      * @throws UserNotFoundException If owner or group members not valid users.
      */
-    void modifyGroup(Group group)
+    Group modifyGroup(Group group)
```
pdowler: What was the purpose of changing this from void to return the modified group instance? 

Also, I see that addGroup is still void so if the change was for consistency/symmetry then that should also be changed (here, in the impl, in the REST API, and in the client).

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapConfig.java
```diff
@@ -90,7 +91,7 @@
 {
     private static final Logger logger = Logger.getLogger(LdapConfig.class);
 
-    public static final String CONFIG = LdapConfig.class.getSimpleName() + ".properties";
+    public static final String CONFIG = "ldap-config.properties";
```
pdowler: Although the naming style is improved, for other services we have implemented a scheme where the service name (ac- in our case) is prepended to the config file name so that ops people can tell which config file(s) go with each deployed service. That change is a little more invasive because you need to propagate the service name (or entire config file name) from a web resource that can access the request (in the case of cadc-access-control-server it is extra painful because there are a bunch of bare servlet impls and no base action class).... maybe create an issue in github for this because it is pretty painful and refactoring the library to use cadc-rest would make it much easier.

For now, change the file name to ac-ldap-config.properties and put a comment beside it that it is a temporary hack.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapConfig.java
```diff
@@ -115,7 +116,8 @@
     public enum PoolPolicy
     {
         roundRobin,
-        fewestConnections
+        fewestConnections,
+        firstResponse
```
pdowler: This must be renamed to fastestConnect (and in the example config) to relfect what it actually does.

PR #65, opened 2019-05-30
S2484: disable /ac/users 'GetUserListAction' function

hjeeves:
To comply with GDPR.  service/ac s2484 branch also has changes to integration tests to support this. If that could be reviewed in conjunction with this pull request, that'd be great!  Thanks.


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/users/UserActionFactory.java
```diff
@@ -108,7 +108,8 @@ public AbstractUserAction createAction(HttpServletRequest request)
 
                 if (segments.length == 0)
                 {
-                    action = new GetUserListAction();
+                    // s2484: GetUserListAction disabled to align with GDPR
```
brianmajor: Just a minor point:  we shouldn't be referencing internal story/task numbers in our open source code because they don't mean anything to our collaborators.

hjeeves: ok. I removed the comment. Thanks!

PR #76, opened 2020-07-13
Story 2738. We now throw a UserAlreadyException when we attempt to a…

yeunga:
…dd a userRequest with either the same user ID or the same email address.

### DIFF
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
@@ -155,31 +155,46 @@ public class LdapUserPersistence extends LdapPersistence implements UserPersiste
     public User addUserRequest(UserRequest userRequest, final Principal ownerHttpPrincipal)
         throws UserNotFoundException, TransientException, AccessControlException, UserAlreadyExistsException
     {
-        LdapUserDAO userDAO = null;
         LdapConnections conns = new LdapConnections(this);
+        LdapUserDAO userDAO = null;
+        LdapGroupDAO groupDAO = null;
         User user = null;
+        Group group = null;
         try
         {
-            // add the userRequest
-            userDAO = new LdapUserDAO(conns);
-            user = userDAO.addUserRequest(userRequest);
-
             // create the group to be associated with this userRequest
-            LdapGroupDAO groupDAO = new LdapGroupDAO(conns, userDAO);
+            userDAO = new LdapUserDAO(conns);
+            groupDAO = new LdapGroupDAO(conns, userDAO);
             LocalAuthority localAuthority = new LocalAuthority();
             URI gmsServiceURI = localAuthority.getServiceURI(Standards.GMS_GROUPS_01.toString());
             GroupURI groupID = new GroupURI(gmsServiceURI, userRequest.getUser().getHttpPrincipal().getName());
-            Group group = new Group(groupID);
+            group = new Group(groupID);
             User groupOwner = userDAO.getAugmentedUser(ownerHttpPrincipal,  false);
             ObjectUtil.setField(group, groupOwner, "owner");
             
+            // add the userRequest
+            user = userDAO.addUserRequest(userRequest);
+           
             // add the user to the group
             group.getUserMembers().add(user);
             
             // add the group associated with the userRequest
             groupDAO.addUserAssociatedGroup(group, user.posixDetails.getGid());
         } catch (GroupAlreadyExistsException ex) {
-            // no need to add the group, do nothing
+            // generate another userRequest
+        	group.getUserMembers().clear();
+            user = userDAO.addUserRequest(userRequest);
+            group.getUserMembers().add(user);
+            
+            try {
+            	// add the group associated with the userRequest
+	            groupDAO.addUserAssociatedGroup(group, user.posixDetails.getGid());
+            } catch (GroupAlreadyExistsException gaex) {
+            	// again a group already exists for a newly created userRequest
+            	// something is wrong
+            	String msg = "BUG: " + ex.getMessage() + ", " + gaex.getMessage();
+                throw new IllegalStateException("BUG: " + ex.getMessage() + ", " + gaex.getMessage());
+            }
         }
         finally
         {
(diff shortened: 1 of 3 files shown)


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
```diff
@@ -160,33 +160,26 @@ public User addUserRequest(UserRequest userRequest, final Principal ownerHttpPri
...
             group.getUserMembers().add(user);
             
             // add the group associated with the userRequest
             groupDAO.addUserAssociatedGroup(group, user.posixDetails.getGid());
         } catch (GroupAlreadyExistsException ex) {
-            // clean up
-            if (user != null) {
-                deleteUserRequest(user.getHttpPrincipal());
-            }
+            // no need to add the group, do nothing
```
brianmajor, line 182: Are you saying that if the group already exists then that's okay?  A group exists for a reason so I don't think we hijack it for this new user...

yeunga: Added code to handle GroupAlreadyExistsException.

brianmajor: It was really just a question about the logic... but I guess you agree that the existence of the group should be checked?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
```diff
@@ -155,37 +155,45 @@ public User addUser(User user)
...
-            Group group = new Group(groupID);
+            group = new Group(groupID);
             User groupOwner = userDAO.getAugmentedUser(ownerHttpPrincipal,  false);
             ObjectUtil.setField(group, groupOwner, "owner");
-
-            // ensure that there is no existing group with the same user name
-            boolean groupExists = checkIfGroupExists(group, groupDAO);
-            if (groupExists) {
-                throw new GroupAlreadyExistsException("Group " + group.getID().getName() + " already exists ");
-            }
```
brianmajor, line 177: Why was this deleted?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
```diff
@@ -155,37 +155,45 @@ public User addUser(User user)
...
             // add the group associated with the userRequest
             groupDAO.addUserAssociatedGroup(group, user.posixDetails.getGid());
         } catch (GroupAlreadyExistsException ex) {
-            // clean up
-            if (user != null) {
-                deleteUserRequest(user.getHttpPrincipal());
+            // generate another userRequest
+        	group.getUserMembers().clear();
+            user = userDAO.addUserRequest(userRequest);
+            group.getUserMembers().add(user);
```
brianmajor, line 187: Why is there a second attempt to create a user request?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
yeunga: Not sure what you mean by 'don't see an update to build.gradle'. The version in build.gradle was updated. Perhaps you mean something else?

The bug was introduced by the deleted lines of code that you mentioned above. GroupDAO.addUserAssociatedGroup() checks whether the group already exists. Instead of checking for the existence of the group, we just create the UserRquest first and then create the group. By doing that, the checks for an already existing UserRequest (due to same userID or same email) fall into place.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
yeunga: When a UserRequest is created, its uid is generated. The same uid is used as the gid. If we detect that the group already exists right after we have just created a UserRequest, it indicates that there is a race condition. If we create another UserRequest (i.e. use a newly generated uid), it's highly unlikely that the group already exists (i.e. a group with the same value as the newly generated uid has just been created). That's the reason for the 2nd attempt to create a user request. I suppose I can synchronize the codes. Just thought that this is simpler and should work most (99.99% ;-) of the time.

yeunga: I incermented build.gradle to 1.1.24 in branch s2738 on July 10. Pat incremened the same build.gradle to 1.1.24 on July 17 and released it to the master branch. This explains why my build.gradle file change did not show up in the PR. I have incremented it again to version 1.1.25.

brianmajor: If there is a collision with a newly created uid/gid, then just fail.  Don't try to do it again.

brianmajor: The fail should be internal server error (500) so a IllegalStateException

yeunga:
> Sorry, I don't see where the bug was fixed. Where are the checks to see if the userid and email address already exist? What exception is thrown in those cases?

The checks are in LdapUserDAO.addUserRequest(). These checks have been there for some time. Both cases throw IllegalArgumentException.

Prior to this fix, the logic checks if a group already exists and deletes the group if it does. This is the bug and it explains why no message was sent to the user.

With this fix, the logic adds the userRequest and then the corresponding group.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/ldap/LdapUserPersistence.java
brianmajor: Making sure you saw these comments...

yeunga: Since Pat has not released his changes, there is no need for me to increment the build version, so I decremented it to keep it at 1.1.24.

Okay. Will update the code to throw an illegalStateException.

yeunga: Done. Please note that there are changes in service.git/ac that need to be reviewed. Basically I added an integration test there.

PR #77, opened 2020-07-17
GroupURI and downstream fixes

pdowler:
also: tomcat realm and LoginServlet fixes to respond 401 in more cases

### DIFF
--- a/cadc-tomcat/src/main/java/ca/nrc/cadc/tomcat/CadcBasicAuthenticator.java
+++ b/cadc-tomcat/src/main/java/ca/nrc/cadc/tomcat/CadcBasicAuthenticator.java
@@ -157,10 +157,9 @@ public class CadcBasicAuthenticator extends RealmBase {
             return null;
         } catch (Throwable t) {
             success = false;
-            String message = "Could not do http basic authentication: "
-                    + t.getMessage();
+            String message = "username/password authentication failed: " + t.getMessage();
             log.error(message, t);
-            throw new IllegalStateException(message, t);
+            return null;
         } finally {
             long duration = System.currentTimeMillis() - start;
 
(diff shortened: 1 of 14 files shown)


cadc-tomcat/src/main/java/ca/nrc/cadc/tomcat/CadcBasicAuthenticator.java
```diff
@@ -157,10 +157,9 @@ public Principal authenticate(String username, String credentials) {
             return null;
         } catch (Throwable t) {
             success = false;
-            String message = "Could not do http basic authentication: "
-                    + t.getMessage();
+            String message = "username/password authentication failed: " + t.getMessage();
             log.error(message, t);
-            throw new IllegalStateException(message, t);
+            return null;
```
brianmajor, line 162: Hmm, I'm wondering if this will be hiding configuration or operational issues.  There's still the stack trace but what do clients see?  A 401 as if they entered the wrong userid/password combo?

pdowler: Yes, it does hide some things from the user and we'll have to detect it in some way that is not "user sees a 500" which is what you get with the IllegalStateException. 

Since I changed LoginServlet to respond 401 when username or password is missing, that is also more obscure (but generally to devs) and either of these alone would hide the 500 from user in the scenario we had. The LoginServlet change will make incorrect use more opaque to developers, but is easy to justify as correct: no password is the same as wrong password... I think the CadcBasicAuthenticator change is arguably technically correct (user is not authenticated) but it does hide more problems than just wrong credentials.

I asked about making one or both of these more opaque in slack, but didn't get any input from ops. Since we cannot control what happens with the throw (the error message sent to the user) I prefer this approach to the tomcat error page and stack traces - those look like bugs to the user.

PR #95, opened 2021-06-19
cadc-9749 Added support for multiple OIDC.

yeunga:
- updated OIDCUtil.getRelyParty(...) to read oidc and corresponding secret from a config file.
- added a sample ac-oidc-clients.properties file.


brianmajor (changes requested):
Could you please add a unit test for the getRelyParty() method?  Then we'll at least have a sample properties files in the opensource code.  The test can be very simple.

yeunga:
> Could you please add a unit test for the getRelyParty() method? Then we'll at least have a sample properties files in the opensource code. The test can be very simple.

Sure. However I have already included a sample properties file.
BTW, the current code does not handle the case when there is no value in the KVP, e.g. 'oidc='. I can add a check for this scenario and throw a RuntimeException.

brianmajor:
Oh, maybe you forgot to add the sample properties file?  I don't think I
saw it in the 'files changed' view.
Don't worry about the missing value case

yeunga:
> Oh, maybe you forgot to add the sample properties file? I don't think I saw it in the 'files changed' view. Don't worry about the missing value case
> […](#)
If expand the first link above on commits, the first file is the ac-oidc-clients.properties.

brianmajor (approved):

PR #101, opened 2021-07-28
CADC-9905, 9909: OIDC login through CADC login page

hjeeves:
1) change LoginAction.java to return a 200 + the redirect url as returning 302 from this service causes web browsers to take control of the redirect after a POST generating a CORS error and blocking the redirect. 
2) change AuthorizeAction.java to return a reference to <cadc host>/en/login.html instead of oidc-login.html.

### DIFF
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
@@ -71,12 +71,14 @@ import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.auth.HttpPrincipal;
 import ca.nrc.cadc.auth.NotAuthenticatedException;
 import ca.nrc.cadc.net.NetUtil;
+import ca.nrc.cadc.reg.client.RegistryClient;
 import ca.nrc.cadc.rest.RestAction;
 
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.net.URL;
 import java.net.URLEncoder;
 import java.util.Arrays;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
```diff
@@ -209,9 +211,18 @@ private void doOpenIDCodeFlow() throws Exception {
         Subject s = AuthenticationUtil.getCurrentSubject();
         AuthMethod authMethod = AuthenticationUtil.getAuthMethodFromCredentials(s);
         if (authMethod.equals(AuthMethod.ANON)) {
-            
+            // Get the service URL for ac
+            RegistryClient regClient = new RegistryClient();
+            URL regCapabilitiesURL = regClient.getAccessURL(new URI("ivo://cadc.nrc.ca/reg"));
+
+            String regCapURLStr = regCapabilitiesURL.toString();
+            String oidcLoginHostURL = regCapURLStr.replace("/reg/capabilities","");
```
brianmajor, line 219: Can you use regCapabilitiesURL.getHost() instead of using replace?  Depending on /reg/capabilities to be there makes it slightly less stable.  You may need to extract the 'scheme' too.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
```diff
@@ -209,9 +211,18 @@ private void doOpenIDCodeFlow() throws Exception {
...
+
             // send redirect to username/password form
-            StringBuilder redirect = new StringBuilder("oidc-login.html#redirect_uri=");
+            // In future, this reference to /en/login.html can be replaced by
+            // a provide an arbitrary login screen.
```
brianmajor, line 223: small grammar error

brianmajor (changes requested):
Looks good, just a few minor comments,

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
hjeeves: Got that. I'd fallen into the pattern of what was done in the javascript. oops!

Not sure about extracting the scheme? The service returns URLs for the appropriate login host on testing.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/AuthorizeAction.java
hjeeves: changed to:
            // In future, this reference to /en/login.html can be replaced by
            // a mechanism to provide an arbitrary login screen.

brianmajor (approved):

PR #118, opened 2022-01-06
CADC-10358 update GroupClient to check for membership in a list of groups

jburke-cadc:


pdowler (commented):
Maybe this wasn't clear: aside from the optimization the cadc-gms GroupClient API should expose the full API specified in the IVOA GMS service spec. That means "search with 0..* group names" and the response is either "all memberships" (if no input groups specified) or "subset the caller belongs to" if groups are specified. So, only cadc-gms:

The spec implies the folllowing method that matches the service API:
```
 public List<GroupURI> getMemberships(List<??> groups);
```
These are really convenience, but probably worthwhile:
```
public List<GroupURI> getMemberships();
public boolean isMember(List<>>> groups)
```
which just returns `!getMemberships(groups).isEmpty()` for callers that don't which group it is. But: the getMemberships(List) above should be part of the API and it's really the main method of the  API (even getMemberships() is a convenience method).

Second issue: I used `List<??> groups` above because I am not sure if it should be `String` or `GroupURI`. The GMS spec assumes the client only asks service X about groups service X is authority for, but the impl you have in cadc-access-control does not do that: it blindly extracts group names from the GroupURI(s) without checking the resourceID -- that's because our impl is bound to a single resourceID in ctor, so that code is wrong. If the type is GroupURI (as you have it now) then a client impl could be written to not be bound to one resourceID and the getMemberships(List<GroupURI>) call could split up the args by reosurceID and call one or more services... if the method signature was getMemberships(List<String>) then they are clearly group names, the client has to be bound to one resourceID, and the caller extracts names and makes sure they use the right client.

So, which is the better API here? I have an opinion, but I"d like see what you think and what you see as the pros/cons/consequences of the API choices.

jburke-cadc:
First part: I suspect I looked at the spec but only thought of the convenience method I needed, because as noted it's a one liner to check if any groups were returned from the missing API method `List<GroupURI> getMemberships(List<??>)`.

API question: I think regardless of whether a single instance of the GMSClient covers one GMS service, or any possible services, the List of groups passed to the `getMemberships(List<??>)` should be GroupURI's. 
If the GMSClient can query any service, you need the GroupURI to extract the services to query. 
If a GMSClient queries a single service, you would want to confirm that groups in the List of groups belong to that service.
If the List of groups was a List of groupNames, you are restricting the client to a single service, since there is no way to know which group belongs to which service.

Looking at it strictly as an API, forgetting about the client, I still think you need to pass in GroupURI's since the spec says the access_url to query comes from the Group serviceID and the standardID.

Should the GMSClient be restricted to a single service via the constructor, less certain. You can't get all memberships for the caller without knowing which service to query. For the other methods you could extract and call the service in each GroupURI parameter.
Is there a use case for the GMSClient querying multiple GMS services? I see `ivo://oats.inaf.it/gms = http://gms01.oats.inaf.it/ac/capabilities` in our resource-caps, which is the Italian VO I believe.  That also implies knowledge whether GroupURI's could be for different services. 

Guess I'm a little confused because the GMS spec is for the web service, but we're defining the spec for the client. And the service spec is pass in groupNames, and return groupNames, not GroupURI's.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -151,51 +153,84 @@ public String getEventHeader()
...
-        if (!group.getServiceID().equals(serviceID)) {
-            throw new UnsupportedOperationException("Group is not in the target GMS service.");
+
+        // discard groups not in the target GMS service
+        List<String> groupNames = new ArrayList<>();
+        for (GroupURI group : groups) {
+            if (group.getServiceID().equals(this.serviceID)) {
+                groupNames.add(group.getName());
+            } else {
+                log.warn(String.format("%s is not in the target GMS service %s",
```
pdowler, line 177: Since we decided "ignore" was the right policy, this should be logged as debug.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
```diff
@@ -771,9 +812,17 @@ private Principal getCurrentUserID()
 
         String roleString = role.getValue();
         URL searchURL = lookupServiceURL(Standards.GMS_SEARCH_01);
-        URL getMembershipsURL = new URL(searchURL.toExternalForm()
-                                        + "?ROLE="
-                                        + NetUtil.encode(roleString));
+        StringBuilder sb = new StringBuilder();
```
pdowler, line 815: The above Standards.GMS_SEARCH_01 is not the correct GMS-1.0 standard ID. The code below uses the old prototype which is obsolete. How does this actually work?

Should be GMS_SEARCH_10; read the spec about output format. You probably need to rewrite this method, in which case the input List<GroupURI> and "ignoring" can happen right in the loop that appends `group` params (line ~820).

pdowler (changes requested):

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
jburke-cadc: Changed from warn to debug.

cadc-access-control/src/main/java/ca/nrc/cadc/ac/client/GMSClient.java
jburke-cadc: My test VM has both Standards: GMS_SEARCH_01 & GMS_SEARCH_10 which is how the int-tests run okay. I'll remove the old Standard.
Yes better that the the group serviceID checking happens in the 'low level' method so the same check doesn't have to happen in multiple places.
The output of that method looks correct, a list of group names the caller belongs to. But it, and several other methods, are using the deprecated HttpDownload instead of HttpGet. Missed that before.

jburke-cadc: Left the `ignore` checking in the public `getMemberships` method as an optimization. If there are no groups left after the ignore check, an empty list can be returned, rather than do the LDAP query, which will just return an empty list.

pdowler:
this is OBE given the new client in cadc-gms

PR #119, opened 2022-01-20
changed TokenTool.validate to accept multiple Grant classes

pdowler:
reason: allows a service to allow a read when a pre-auth token has a WriteGrant; needed in minoc to allow a head (read) after a put (with write grant pre-auth) both within and after a transaction.

I thought about more explicitly changing WriteGrant to ReadWriteGrant but that would be more invasise right now.

### DIFF
--- a/cadc-permissions/build.gradle
+++ b/cadc-permissions/build.gradle
@@ -14,7 +14,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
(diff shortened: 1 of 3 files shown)


brianmajor (approved):
Yes, as you mention I think changing WriteGrant to ReadWriteGrant at some point would be an improvement and would also match the patterns we use elsewhere.

PR #122, opened 2022-02-25
fix custom comparator use for java 11 and Subject null handling

pdowler:


brianmajor (approved):
Learned something new today.

PR #123, opened 2022-03-03
Cadc 10630 Updated LoginServlet to prevent incorrect username in tokens

yeunga:
Added logic in LoginServlet to remove the principal created from a userID with different cases than the one in ldap.

### DIFF
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
@@ -152,13 +152,7 @@ public class AuthenticatorImpl implements Authenticator
 
             // CADC-10630 Remove potentially incorrect userID
             // in HttpPrincipal in subject.
-            Set<Principal> toBeRemoved = new HashSet<Principal>();
-            subject.getPrincipals().forEach(p -> {
-                if (p instanceof HttpPrincipal) {
-                    toBeRemoved.add(p);
-                }
-            });
-            subject.getPrincipals().removeAll(toBeRemoved);
+            subject.getPrincipals().removeAll(subject.getPrincipals(HttpPrincipal.class));
 
             User user = userPersistence.getAugmentedUser(ldapPrincipal, true);
             if (user.getIdentities() != null)
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
@@ -80,8 +80,6 @@ import ca.nrc.cadc.profiler.Profiler;
 
 import java.security.AccessControlException;
 import java.security.Principal;
-import java.util.Iterator;
-import java.util.Set;
 
 import javax.security.auth.Subject;
 import javax.security.auth.x500.X500Principal;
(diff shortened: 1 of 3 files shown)


cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
```diff
@@ -147,6 +149,21 @@ public void augmentSubject(final Subject subject)
...
+                Principal princ = iterator.next();
+                if (princ instanceof HttpPrincipal) {
+                    toBeRemoved = princ;
+                    break;
+                }
+            }
+            if (toBeRemoved != null) {
+                subject.getPrincipals().remove(toBeRemoved);
+            }
+
```
brianmajor, line 166: Like we discussed, I think all you need to do is, at this same place in the code:
`subject.getPrincipals().clear()`

brianmajor (changes requested):

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
brianmajor: Actually, only removing the HttpPrincipals would be safer.  So:
`subject.getPrincipals(HttpPrincipal.class).clear()`

yeunga: Right. Thanks. Done.

brianmajor (dismissed):
Looks good.

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
```diff
@@ -147,6 +149,18 @@ public void augmentSubject(final Subject subject)
             PluginFactory pluginFactory = new PluginFactory();
             UserPersistence userPersistence = pluginFactory.createUserPersistence();
             Principal ldapPrincipal = getLdapPrincipal(subject);
+
+            // Remove the HttpPrincipal in subject, if any
```
brianmajor, line 153: Could you please add a comment as to why the http principal is being removed here?

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
```diff
@@ -147,6 +149,18 @@ public void augmentSubject(final Subject subject)
...
+            Set<Principal> toBeRemoved = new HashSet<Principal>();
+            subject.getPrincipals().forEach(p -> {
+                if (p instanceof HttpPrincipal) {
+                    toBeRemoved.add(p);
+                }
+            });
+            toBeRemoved.forEach(p -> {
+                subject.getPrincipals().remove(p);
+            });
+
```
brianmajor, line 163: Would `removeAll(toBeRemoved)` be easier to read here?

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
```diff
@@ -147,6 +149,18 @@ public void augmentSubject(final Subject subject)
             PluginFactory pluginFactory = new PluginFactory();
             UserPersistence userPersistence = pluginFactory.createUserPersistence();
             Principal ldapPrincipal = getLdapPrincipal(subject);
+
+            // Remove the HttpPrincipal in subject, if any
+            Set<Principal> toBeRemoved = new HashSet<Principal>();
```
brianmajor, line 154: Does this work?  `toBeRemoved = subject.getPrincipals(HttpPrincipal.class)`

brianmajor (changes requested):

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
yeunga: Good idea. This does work.

cadc-access-control-server/src/main/java/ca/nrc/cadc/auth/AuthenticatorImpl.java
yeunga: Great idea! It does. Now we are back to a one liner.

brianmajor (approved):

PR #131, opened 2022-06-30
CADC-10684: add kid header to JWT so OIDC validator can run

hjeeves:
Even though this is an optional header, the website we're using to test the quality of our JWTs seems to require it. kid (Key ID) is now set up to be a well known string, but encoded so it's not immediately readable.  (Code in OIDCUtil)

Same value is generated in the JWKS endpoint (code in PublicKeyAction.)

### DIFF
--- a/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/OIDCUtil.java
+++ b/cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/OIDCUtil.java
@@ -71,14 +71,11 @@ import ca.nrc.cadc.ac.GroupNotFoundException;
 import ca.nrc.cadc.ac.Role;
 import ca.nrc.cadc.ac.User;
 import ca.nrc.cadc.ac.UserNotFoundException;
-import ca.nrc.cadc.ac.client.GroupMemberships;
 import ca.nrc.cadc.ac.server.GroupPersistence;
 import ca.nrc.cadc.ac.server.UserPersistence;
 import ca.nrc.cadc.ac.server.ldap.LdapGroupPersistence;
-import ca.nrc.cadc.ac.server.ldap.LdapPersistence;
 import ca.nrc.cadc.ac.server.ldap.LdapUserPersistence;
 import ca.nrc.cadc.ac.server.oidc.RelyParty.Claim;
-import ca.nrc.cadc.auth.AuthMethod;
 import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.auth.HttpPrincipal;
 import ca.nrc.cadc.auth.NumericPrincipal;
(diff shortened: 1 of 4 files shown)


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/OIDCUtil.java
```diff
@@ -415,5 +412,15 @@ static String getClaimDescriptionString(List<Claim> claims) {
...
+
+    public static String getEncodedKID() {
+        return Base64.getUrlEncoder().encodeToString(KID_CLAIM_VALUE.getBytes());
+    }
+
+    public static boolean isRightKIDKey(String encodedKID) {
+        byte[] urlDecodedKIDbytes = Base64.getUrlDecoder().decode(encodedKID.getBytes());
+        String urlDecodedKID = new String(urlDecodedKIDbytes);
+        return KID_CLAIM_VALUE.equals(urlDecodedKID);
+    }
```
brianmajor, line 424: This method does not seem to be used.  Delete?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/OIDCUtil.java
```diff
@@ -415,5 +412,15 @@ static String getClaimDescriptionString(List<Claim> claims) {
         }
         return sb.toString();
     }
+
+    public static String getEncodedKID() {
+        return Base64.getUrlEncoder().encodeToString(KID_CLAIM_VALUE.getBytes());
```
brianmajor, line 417: I wonder if this should use a one-way encoder so that clients can't read the contents... Did you see any recommendations on that?

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/RelyParty.java
```diff
@@ -85,6 +85,7 @@ public class RelyParty {
     
     private String clientID;
     private String clientSecret;
+    private String kidVersion;
```
brianmajor, line 88: What is this field (and the getter and setter) used for?

brianmajor (changes requested):

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/RelyParty.java
brianmajor: This should be removed because the key is not related to a rely part at all.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/oidc/OIDCUtil.java
hjeeves: As discussed in slack - the KID_CLAIM_VALUE is going to be a simple string unrelated to the file name for the RSA priv/pub keys, not encoded.

brianmajor (approved):

PR #132, opened 2022-10-13
CADC-11904: add debugging log msgs to TokenTool 

hjeeves:
So debugging issues with tokens will be easier in future.

### DIFF
--- a/cadc-permissions/src/main/java/org/opencadc/permissions/TokenTool.java
+++ b/cadc-permissions/src/main/java/org/opencadc/permissions/TokenTool.java
@@ -138,7 +138,11 @@ public class TokenTool {
         if (privateKey == null) {
             throw new IllegalStateException("cannot generate token: no private key");
         }
-        
+
+        log.debug("[TokenTool.generateToken]: uri: " + uri);
+        log.debug("[TokenTool.generateToken]: grant: " + grantClass);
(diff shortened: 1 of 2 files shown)


brianmajor (changes requested):
version should probably just be 0.3.4 instead of 0.4.  otherwise looking good.

cadc-permissions/src/main/java/org/opencadc/permissions/TokenTool.java
```diff
@@ -238,19 +242,21 @@ public String validateToken(String token, URI expectedURI, Class<? extends Grant
                 user = value;
             }
         }
-        log.debug("uri: " + uri);
-        log.debug("grant: " + grant);
-        log.debug("subject: " + user);
-        
+        log.debug("[TokenTool.validateToken]: uri: " + uri);
```
pdowler, line 245: Having to put the TokenTool class name here is a sign of our generally poor logging config, where we only log the message (info level start and end messages are json) so we lose out on all the stuff log4j does for us :-(

and to top it off, the json makes logs unreadable and we don't ingest them into something that can search anyway :-( :-(

pdowler: This could have also been fixed more widely by fixing the Log4jInit to use a proper log format for debug messages

pdowler (approved):

PR #134, opened 2022-11-10
prototype IVOA group client

pdowler:


brianmajor (commented):
The public API looks good

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -0,0 +1,272 @@
...
+            }
+            NumericPrincipal p = new NumericPrincipal(uuid);
+            
+            // effectively augment by using the current subject as a "cache" of known identities
+            Subject s = AuthenticationUtil.getCurrentSubject();
+            if (s != null) {
+                for (Principal cp : s.getPrincipals()) {
+                    if (AuthenticationUtil.equals(p, cp)) {
+                        log.debug("[cache hit] caller Subject matches " + p + ": " + s);
+                        ret.getPrincipals().addAll(s.getPrincipals());
```
brianmajor, line 157: I guess all the principals are added because if we match on any (check on line 155), then we can trust that a previous augment subject was done correctly.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -0,0 +1,272 @@
...
+            return ps.iterator().next().getName(); // kind of ugh
+        }
+        
+        // try X509
+        Set<X500Principal> px = subject.getPrincipals(X500Principal.class);
+        if (!px.isEmpty()) {
+            return px.iterator().next().getName(); // kind of ugh
+        }
+        
+        return null;
```
brianmajor, line 194: Don't want to use getName() of whatever else might be in the subject (instead of null)?

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -0,0 +1,272 @@
...
+                } // else: some other challenge 
+                log.debug("challenge type: " + challengeType);
+                log.debug("credentials: " + credentials);
+                    
+                // validate
+                if (challengeType != null && credentials != null) {
+                    try {
+                        HttpGet get = new HttpGet(u, true);
+                        get.setRequestProperty("authorization", raw.getHeaderValue());
+                        get.prepare();
```
brianmajor, line 229: Just want to confirm that prepare() will throw an exception for a non-200 response code

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -0,0 +1,272 @@
...
+
+                        InputStream istream = get.getInputStream();
+                        String str = StringUtil.readFromInputStream(istream, "UTF-8");
+                        JSONObject json = new JSONObject(str);
+                        String sub = json.getString("sub");
+                        String username = json.getString("preferred_username");
+                        // TODO: register an X509 DN with IAM and see if I can get it back here
+
+                        NumericPrincipal np = new NumericPrincipal(UUID.fromString(sub));
```
brianmajor, line 238: I wonder if it's okay to presume a UUID as the subject

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
pdowler: yes, prepare() throws all the exceptions

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
pdowler: yes, `sub` is structurally a UUID, at least with IAM, although OIDC includes language in the spec that only issuer + sub is sure to be unique so it might not be a real (global) UUID. It's necessary to use NumericPrincipal. I think in future we could support int/long uids (like we do in AC) with a little more code, if necessary (eg if there is an OID provider impl that doesn't use UUID in sub).

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
pdowler: Aside from HttpPrincipal or X500, which other one can there actually be? Don't think one wants to output NumericPrincipal here... this is for logging,  x-vo-authenticated header output, job ownerID field output, etc.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
pdowler: yup; unlike ACIdentityManager I didn't just return the cached subject (because it has other junk in it) so copied the principals instead.

brianmajor (approved):

pdowler:
Any comment  on renaming to OidcIdentityManager? or something else?

The idea form the current name was "supports standard IM" (vs custom like AC)

PR #138, opened 2023-03-03
Cadc 1050 Replaced usage of LocalAuthority with a base URL obtained from request URL

yeunga:


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/WhoAmIServlet.java
```diff
@@ -206,15 +201,12 @@ public AuthMethod getAuthMethod(Subject subject)
...
     {
-        final RegistryClient registryClient = getRegistryClient();
-
-        URI umsServiceURI = getServiceURI(Standards.UMS_WHOAMI_01);
-        log.debug("ums service uri: " + umsServiceURI);
-
-        final URL serviceURL = registryClient.getServiceURL(umsServiceURI, Standards.UMS_USERS_01, authMethod);
-        final URL redirectURL = new URL(serviceURL.toExternalForm() + USER_GET_PATH);
+        // TODO: Look for a flexible way (vs currently hardcoded) to append '/users' to the baseServiceURL
+        String baseServiceURL = requestURL.substring(0, requestURL.lastIndexOf("/"));
```
pdowler, line 207: This might be crazy, but what about simply replacing WHOAMI with USERS since both are constants? That should work even if the service is renamed (see main comment).

pdowler (changes requested):
Ideally, service code doesn't depend on how it is deployed (war name or path). You can test this (manually to make sure your implementation works well) with the container deployment by creating a file /config/war-rename.conf:
```
mv ac.war foo.war
```
and calling, eg: /foo/whoami, which should redirect to /foo/users/...

Also, try `mv ac.war foo#bar.war`, which deploys as /foo/bar (so you also have to fix haproxy rules) and then try `/foo/bar/whoami`.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/EndpointConstants.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * This class defines the constants for the strings that conform to the IVOA
+ * service standard identifiers.
+ *
+ * @author yeunga
+ */
+public class EndpointConstants {
+
+    // standardID values
+    public static final String CAPABILITIES = "/capabilities";
```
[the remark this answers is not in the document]
yeunga: Removed the first 3 constants.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/WhoAmIServlet.java
yeunga: Replaced "WHOAMI" constant with "USERS".

yeunga:
I'll look into how to deploy ac as a container and do the manual test above.

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/WhoAmIServlet.java
pdowler: oops, what I meant was that finding the base by looking for the last `/` is potentially fragile; I was suggesting something like
```
String usersEndpoint = requestURL.toString().replace(EndpointConstants.WHOAMI, EndpointConstants.USERS);
```
since /whoami and /users are always siblings. However, you'd also have to be careful to strip off any cruft from the end (bogus query string or fragments, etc) so doing the replace will work in the correct usage but potentially not if someone is messing around.

instead of looking for the `lastIndexOf("/") you could look for the `firstIndexOf(EndpointConstants.WHOAMI)` as a more robust indicator of the end of the base service URL. Maybe `firstIndexOf("/" + EndpointConstants.WHOAMI)` depending to get the same index as before...

yeunga: My misunderstanding. Using first index now.

yeunga:
Verified that the code worked with `/foo/whoami `and `/foo/bar/whoami.`

PR #140, opened 2023-03-21
Cadc 12340 -improve AC container deployment

yeunga:
- stop using dbrc and configure credentials directly in ac-ldap-config.properties
- get oidc and rsa pub/priv keys from config dir

Not ac integration tested

### DIFF
--- a/cadc-access-control-server/build.gradle
+++ b/cadc-access-control-server/build.gradle
@@ -35,8 +35,6 @@ dependencies {
     compile 'org.opencadc:cadc-uws:[1.0,)'
     compile 'org.opencadc:cadc-uws-server:[1.2.4,)'
 
-    runtime 'org.opencadc:cadc-registry:[1.6,2.0)'
-
     testCompile 'junit:junit:[4.0,)'
     testCompile 'org.easymock:easymock:3.6'
     testCompile 'org.skyscreamer:jsonassert:[1.0,)'
--- a/cadc-access-control-server/build.gradle
+++ b/cadc-access-control-server/build.gradle
@@ -13,7 +13,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.3.29'
+version = '1.3.30'
 
 description = 'OpenCADC User+Group server library'
 def git_url = 'https://github.com/opencadc/ac'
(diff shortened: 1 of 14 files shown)


cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/EndpointConstants.java
```diff
@@ -0,0 +1,89 @@
...
+ * service standard identifiers.
+ *
+ * @author yeunga
+ */
+public class EndpointConstants {
+
+    // standardID values
+    public static final String USERS = "/users";
+    public static final String USER_REQUESTS = "/userRequests";
+    public static final String REQUEST_PASSWORD = "/requestPassword";
```
pdowler, line 83: resetPassword?

cadc-access-control-server/build.gradle
```diff
@@ -32,10 +32,11 @@ dependencies {
     compile 'org.opencadc:cadc-access-control:[1.1.23,2.0)'
     compile 'org.opencadc:cadc-util:[1.9.2,)'
     compile 'org.opencadc:cadc-log:[1.1.0,)'
-    compile 'org.opencadc:cadc-registry:[1.6,2.0)'
     compile 'org.opencadc:cadc-uws:[1.0,)'
     compile 'org.opencadc:cadc-uws-server:[1.2.4,)'
 
+    runtime 'org.opencadc:cadc-registry:[1.6,2.0)'
```
pdowler, line 38: seems odd to declare a runtime dependency in a library... should only be needed for a lib that provides a plugin not needed to compile

pdowler (changes requested):

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/EndpointConstants.java
yeunga: Added resetPassword.

cadc-access-control-server/build.gradle
yeunga: Removed line 38.

pdowler (approved):

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/EndpointConstants.java
brianmajor: Should REQUEST_PASSWORD be RESET_PASSWORD?

yeunga: Don't remember where I came up with /requestPassword. Don't seem to be able to find it either. If it's an extra definition, it can be deleted.

Issue #142, opened 2023-05-25
LdapConfig still required dbrcHost entry in config


pdowler:
even though it is not used

pdowler:
see PR #143

PR #143, opened 2023-06-07
cadc-gms: update GroupURI to allow / (hierarchical) group names; resolves #139 and #142

### LINKED ISSUE #139: Token scope needs to expand , currently it is limited to domain of its issuer.
When calling underlying apis from storage inventory the Authorization token is getting passed only when domain of the issuer of token is same as the calling api.
 the logic to setAuthorization header is present in cadc-util (ca.nrc.cadc.net.HttpTransfer)

      for (String domain : next.getDomains()) { 
          if (conn.getURL().getHost().endsWith(domain)) {
          ...............
          }} 
  
 Storage inventory can be deployed into different regions and the domain might be different for those regions , in this case Authorization token is not getting passed. This logic should be changed to accommodate all the domains.

pdowler:
update StandardIdentityManager to add a trusted GMS host to oidcDomains 
deprecate old GroupCLient interface

### DIFF
--- a/cadc-gms/src/main/java/org/opencadc/gms/GroupClient.java
+++ b/cadc-gms/src/main/java/org/opencadc/gms/GroupClient.java
@@ -79,6 +79,7 @@ import java.util.List;
  * @author majorb
  *
  */
+@Deprecated
 public interface GroupClient {
     
     /**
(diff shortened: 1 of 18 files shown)


cadc-gms/src/main/java/org/opencadc/gms/GroupClient.java
```diff
@@ -79,6 +79,7 @@
  * @author majorb
  *
  */
+@Deprecated
```
brianmajor, line 82: Comment here as to the reason and alternatives to this now deprecated class?

brianmajor (approved):

PR #149, opened 2023-09-15
cadc-gms: add stub PosixMapperClient to auth package

pdowler:
also tweak cadc-access-control impls to return minimal posix details in augment subject

### DIFF
--- a/cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
+++ b/cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
@@ -1,0 +1,144 @@
+/*
+************************************************************************
+*******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
+**************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
+*
+*  (c) 2023.                            (c) 2023.
+*  Government of Canada                 Gouvernement du Canada
+*  National Research Council            Conseil national de recherches
+*  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
+*  All rights reserved                  Tous droits réservés
+*
+*  NRC disclaims any warranties,        Le CNRC dénie toute garantie
+*  expressed, implied, or               énoncée, implicite ou légale,
+*  statutory, of any kind with          de quelque nature que ce
+*  respect to the software,             soit, concernant le logiciel,
+*  including without limitation         y compris sans restriction
+*  any warranty of merchantability      toute garantie de valeur
+*  or fitness for a particular          marchande ou de pertinence
+*  purpose. NRC shall not be            pour un usage particulier.
+*  liable in any event for any          Le CNRC ne pourra en aucun cas
+*  damages, whether direct or           être tenu responsable de tout
+*  indirect, special or general,        dommage, direct ou indirect,
+*  consequential or incidental,         particulier ou général,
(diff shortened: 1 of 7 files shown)


cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+    
+    // use case: skaha needs the complete groupname-gid map for user containers
+    public Iterator<PosixGroup> getGroupMap() {
+        throw new UnsupportedOperationException();
+    }
+    
+    // use case: skaha getGroupMap response - probably extracted into a normal class
+    // depending on where/how we deal with user default group
+    public static class PosixGroup {
```
at88mph, line 133: If all goes according to plan, then the Group entry would also have the members:

`CHIMERWX:x:9999:jenkinsd,adamian,sgaudet`

cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+    
+    // use case: cavern needs PosixPrincipal in the subject
+    // use case: StandardIdentityManager calls this to add local identity
+    public void augment(Subject s) {
+        throw new UnsupportedOperationException();
+    }
+    
+    // use case: skaha needs to complete username-uid map for user containers
+    public Iterator<PosixPrincipal> getUserMap() {
```
at88mph, line 122: The `PosixPrincipal` only contains a UID.  For a complete mapping, this would need to include the Name, Home Directory, and default Shell, which could just default to `/sbin/nologin`.  Is that the purpose of this method?

cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
pdowler: This is supposed to just be the mapping of name <-> gid... does it really need the list of members here? That won't come from the mapper service - it comes from a GMS service. But it is a good point... it changes the complexion of the injected db a lot. Need to discuss.

cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
pdowler: I'm adding username and defaultGroup to PosixPrincipal (cavern only needs the default group). In principle posix username is local and thus could differ from "network" username (HttpPrincipal) so I'm pretty sure it should be there but we don't know enough about the possibilities there.

yes, the purpose is to provide a local uid<->username mapping to construct a "passwd" file... I'm hoping minimal user info can reside solely in the Subject and that looks plausible, but we might need to add more fields to `PosixPrincipal`

pdowler: see https://github.com/opencadc/core/pull/230 for the PosixPrincipal change (so far)

see other changes in this PR to make the AC augment subject calls get those posix values to the client (build failing until cadc-util is published)

cadc-gms/src/main/java/org/opencadc/auth/PosixMapperClient.java
pdowler: moving discussion to `team-src-red` slcack channel

at88mph (approved):
For Skaha and User Sessions, this works for me.

PR #151, opened 2023-09-21
POSIX Mapping service

at88mph:
Map UID and Usernames
Map GID and Group Names
Create Groups on the fly with `group=<URI>`


posix-mapper/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+
+war {
+    from(System.getenv('RPS') + '/resources/') {
+        include 'swagger-ui/'
+    }
+    from('.') {
+        include 'VERSION'
+    }
+
+    archiveName 'posix.war'
```
pdowler, line 26: in every other project we use the default project name for the war (eg posix-mapper). This will be slightly odd for no reason. Please use default and fix Dockerfile and README

posix-mapper/src/intTest/java/org/opencadc/posix/mapper/GroupManagementIntTest.java
```diff
@@ -0,0 +1,181 @@
...
+    public static final URI POSIX_MAPPER_SERVICE_ID = URI.create("ivo://opencadc.org/posix-mapper");
+
+    static {
+        Log4jInit.setLevel("org.opencadc.posix.mapper", Level.DEBUG);
+    }
+
+    private static final String TEXT_PLAIN_CONTENT_PLAIN = "text/plain";
+    private static final String TSV_CONTENT_PLAIN = "text/tab-separated-values";
+    private static final URI GROUP_MAPPER_STANDARD_ID =
+            URI.create("http://www.opencadc.org/std/posix#group-mapping-1.0");
```
pdowler, line 106: This should use the constant from Standards (cadc-registry)

posix-mapper/src/intTest/java/org/opencadc/posix/mapper/UserManagementIntTest.java
```diff
@@ -0,0 +1,194 @@
...
+    private static final Logger log = Logger.getLogger(UserManagementIntTest.class);
+    public static final URI POSIX_MAPPER_SERVICE_ID = URI.create("ivo://opencadc.org/posix-mapper");
+
+    static {
+        Log4jInit.setLevel("org.opencadc.posix.mapper", Level.DEBUG);
+    }
+
+    private static final String TEXT_PLAIN_CONTENT_PLAIN = "text/plain";
+    private static final String TSV_CONTENT_PLAIN = "text/tab-separated-values";
+    private static final URI USER_MAPPER_STANDARD_ID = URI.create("http://www.opencadc.org/std/posix#user-mapping-1.0");
```
pdowler, line 109: constant from Standards class

posix-mapper/src/intTest/java/org/opencadc/posix/mapper/UserManagementIntTest.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        return Integer.parseInt(byteArrayOutputStream.toString());
+    }
+
+    private void getUsers(final OutputStream outputStream, final String contentType) throws Throwable {
+        final InputStreamWrapper inputStreamWrapper = inputStream -> outputStream.write(inputStream.readAllBytes());
+        final HttpGet httpGet = new HttpGet(userMapperURL, inputStreamWrapper);
+        httpGet.setRequestProperty("accept", contentType);
+        httpGet.run();
```
pdowler, line 188: HttpTransfer(s) support a new more direct mode:

instead of calling `run()` you can call `prepare()` which does everything except read the stream and throws a variety of exceptions; then you don't haveto check `getThrowable()`. If `prepare()` returns (success) then you can call `getInputStream()` and check it or (like here) just close it).

posix-mapper/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,150 @@
...
+        <servlet-class>ca.nrc.cadc.rest.RestServlet</servlet-class>
+        <init-param>
+            <param-name>augmentSubject</param-name>
+            <param-value>false</param-value>
+        </init-param>
+        <init-param>
+            <param-name>init</param-name>
+            <param-value>org.opencadc.posix.mapper.web.PosixInitAction</param-value>
+        </init-param>
+        <init-param>
```
pdowler, line 47: I don't understand why there is a post action? doesn't get action create a mapping if necessary and return it? That was the idea (for both user and group)

posix-mapper/src/main/webapp/capabilities.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<vosi:capabilities
+    xmlns:vosi="http://www.ivoa.net/xml/VOSICapabilities/v1.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:vs="http://www.ivoa.net/xml/VODataService/v1.1">
+
+  <capability standardID="http://www.opencadc.org/std/posix#user-mapping-1.0">
```
pdowler, line 7: the constants in Standards use version 0.1 for prototype so that should be used here (and below)

posix-mapper/src/main/webapp/META-INF/context.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <WatchedResource>WEB-INF/web.xml</WatchedResource>
+
+    <Resource name="jdbc/posix-mapper"
+        auth="Container"
+        type="javax.sql.DataSource"
+        factory="org.apache.tomcat.jdbc.pool.DataSourceFactory" closeMethod="close"
+        minEvictableIdleTimeMillis="60000" timeBetweenEvictionRunsMillis="30000"
+        maxWait="20000"
+        initialSize="0" minIdle="0" maxIdle="${org.opencadc.science-platform.mapping.maxActive}" maxActive="${org.opencadc.science-platform.mapping.maxActive}"
```
pdowler, line 12: The keys here and in the README (catalina.properties) should match the java package, so
```
org.opencadc.posix.mapper.???
```
Ah, the README does say that so the fix is just here.

pdowler (changes requested):
General comment: the rest action layer: I was able to implement this in `ac`) with two classes of about 30 lines of code, all in a single doAction method. You have 3 packages and 10 classes to do almost the same thing. It's not the amount of code, but the complexity to figure out how to make a small change 6 months from now. Nothing to do about it, just a comment that RestAction makes it really easy to do simple things and minimise complexity.

posix-mapper/src/main/java/org/opencadc/posix/mapper/db/InitializeMappingDatabase.java
```diff
@@ -0,0 +1,119 @@
...
+        super(dataSource, database, schema, InitializeMappingDatabase.MODEL_NAME,
+              InitializeMappingDatabase.MODEL_VERSION, InitializeMappingDatabase.PREV_MODEL_VERSION);
+
+        this.createSQL.addAll(Arrays.stream(InitializeMappingDatabase.CREATE_SQL).collect(Collectors.toList()));
+    }
+
+    @Override
+    public List<String> parseDDL(String fileName, String schema, String tag) throws IOException {
+        final MultiValuedProperties configuration = PosixInitAction.getConfig();
+        return super.parseDDL(fileName, schema, tag).stream()
```
pdowler, line 107: ah, clever. This will of course mean that one can only set the min values once rather than change them later with `alter sequence`... but that's probably fine for now.

posix-mapper/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM images.opencadc.org/library/cadc-tomcat:1
+
+COPY build/libs/posix-mapper.war /usr/share/tomcat/webapps/posix.war
```
pdowler, line 3: The resulting war name needs to also be posix-mapper.war so that people can correctly know how to use `war-rename.conf` if they decide they want to.

pdowler (changes requested):

posix-mapper/Dockerfile
at88mph: Ah, I see.  I missed that.

pdowler (approved):

PR #154, opened 2023-09-26
cadc-gms: StandardIdentityManager uses PosixMapperClient for extra local augment

pdowler:


cadc-gms/src/intTest/resources/config/cadc-registry.properties
```diff
@@ -0,0 +1,10 @@
+## really temporary intTest setup so pdowler can check that code still works
+## NOT expected to work for anyone else
+ca.nrc.cadc.reg.client.RegistryClient.baseURL = https://haproxy.cadc.dao.nrc.ca/reg
+
+# configure LocalAuthority lookups
+## SRC IAM prototype 
+ivo://ivoa.net/sso#OpenID = https://ska-iam.stfc.ac.uk/                                                                                       
+
+http://www.opencadc.org/std/posix#group-mapping-0.1 = ivo://opencadc.org/src/posix-mapper
```
at88mph, line 9: Is the `user-mapping-0.1` intentionally missing here?  The `StandardIdentityManager` changes below only look for that, never the `group-mapping-0.1`.

at88mph (changes requested):
Just one clarification on how this is used.

at88mph (approved):

at88mph:
Oh, this needs a version increment in the `build.gradle` file as well.

cadc-gms/src/intTest/resources/config/cadc-registry.properties
pdowler: strange - don't know how that test managed to pass because it should be user-mapping-0.1

fixing

pdowler: I will fix the intTest to find the test token in classpath later

PR #159, opened 2023-10-18
Implement UserMap and GroupMap output.

### LINKED ISSUE #158: Implement GroupMap and UserMap in PosixMapperClient
This is needed by Skaha as it will pass it to the Job Creation.

at88mph:
Fixes #158

### DIFF
--- a/cadc-access-control-server/build.gradle
+++ b/cadc-access-control-server/build.gradle
@@ -13,7 +13,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.3.34'
+version = '1.3.33'
 
 description = 'OpenCADC User+Group server library'
 def git_url = 'https://github.com/opencadc/ac'
--- a/cadc-access-control-server/build.gradle
+++ b/cadc-access-control-server/build.gradle
@@ -13,7 +13,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.3.33'
+version = '1.3.34'
 
 description = 'OpenCADC User+Group server library'
 def git_url = 'https://github.com/opencadc/ac'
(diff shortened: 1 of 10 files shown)


pdowler (commented):
for PosixMapperClient, you can already get the whole list by querying with an empty arg list so I thought we considered those iterator methods to be convenience that we didn't need (because the init container would just hit the URL itself and not involve java code). What's the use cases for adding this other than iterator is more scalable? 

aside: the CADC/CANFAR uidmap is 250KiB and the gidmap is 375KiB

as for the other change, I don't know that `preferred_username` is really spec compliant. That is one of the discrepancies between IAM and Rubin/LSST, which also claim OIDC complaince. So I'm not sure what that is fixing exactly... is it supposed to make `ac` more OIDC compliant than it is?

These things sometimes need to be discussed before working on them

at88mph:
The comments of the `getUserMap` and `getGroupsMap` stated the exact use case that Skaha has, so it appeared as though it just wasn't complete, especially since Skaha _used_ to just call the endpoint directly, but no longer does.  It now relies on the API of the `PosixMapperClient`.  Also, there was no previous way to get all of the `PosixPrincipal` instances that I could see.

As for adding the `preferred_username`, that's stated in the `StandardIdentityManager`:
https://github.com/opencadc/ac/blob/master/cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java#L329

Plus it lines up with the SKAO IAM service, and what openid.net suggests:
https://openid.net/specs/openid-connect-basic-1_0.html#Scopes

pdowler (changes requested):
The PosixMapperClient work looks fine, but I should have defined the API using ResourceIterator instead of Iterator (so the underlying input stream close() gets called).

Can you make that improvement and remove the changes from cadc-access-control-server?

at88mph:
Alright, that's complete now.

PR #165, opened 2023-12-12
Instantiate TokenTool with in memory keys (byte arrays) (CADC-12974)

andamian:


cadc-permissions/src/test/java/org/opencadc/permissions/TokenToolTest.java
```diff
@@ -140,15 +144,33 @@ public void testRoundTripToken() {
                 "C=CA, O=Grid, OU=nrc-cnrc.gc.ca, CN=Brian Major",
                 "C=CA, O=Grid, OU=nrc-cnrc.gc.ca, CN=Brian Major",};
 
+            // keys in file
             TokenTool gen = new TokenTool(pubKeyFile, privateKeyFile);
             TokenTool ver = new TokenTool(pubKeyFile);
+            // keys in memory
+            KeyPairGenerator kpg;
```
pdowler, line 151: this setup code (lines 151-160) would be replaced by call to `RsaSignatureGenerator.generateKeyPair(int)`.

pdowler (changes requested):

cadc-permissions/src/test/java/org/opencadc/permissions/TokenToolTest.java
andamian: Yeah. I thought it might be too simple to bother with.

pdowler (approved):

PR #166, opened 2024-02-20
bug fix: detect required augment more carefully

pdowler:
this has the downside that it will cause augment subject to be done to fill in missing PosixPrincipal fields everywhere when in fact only cavern and skaha care about PosixPrincipal details

might need a way to mitigate this (at least temporarily) -- enable posix augment where needed

### DIFF
--- a/cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
+++ b/cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
@@ -75,6 +75,7 @@ import ca.nrc.cadc.auth.HttpPrincipal;
 import ca.nrc.cadc.auth.IdentityManager;
 import ca.nrc.cadc.auth.NotAuthenticatedException;
 import ca.nrc.cadc.auth.NumericPrincipal;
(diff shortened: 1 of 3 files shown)


cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
```diff
@@ -132,12 +133,17 @@ public Subject augment(final Subject subject) {
...
-        // AND the list is greater than 1, then subject has already been augmented
-        Set<Principal> principalSet = subject.getPrincipals();
-        Set<NumericPrincipal> nps = subject.getPrincipals(NumericPrincipal.class);
-        if (principalSet.size() > 1 && !nps.isEmpty()) {
+        NumericPrincipal np = getNumericPrincipal(subject);
+        boolean needAugment = np == null;
+
+        PosixPrincipal pp = getPosixPrincipal(subject);
+        needAugment = needAugment || (pp == null || pp.defaultGroup != null || pp.username != null); // missing or incomplete
+
```
brianmajor, line 145: Just the logic change here to be == instead of !=

brianmajor (changes requested):

brianmajor (dismissed):

PR #169, opened 2024-04-25
Proposal for predefined key access

at88mph:
Allow setting authorized Groups (empty list means nobody is allowed) in POSIX Mapper
Add reusable pre-defined key checker to access POSIX Mapper


at88mph:
The POSIX Mapper can now be used with a request header:
`authorization: api-key {TOKEN}`

Groups are also checked for `Bearer` tokens of authorized users.

Supports checking key files in the `${HOME}/config/keys/` folder.

posix-mapper/docker/config/posix-mapper/catalina.properties
```diff
@@ -0,0 +1,13 @@
+tomcat.connector.scheme=https
```
pdowler, line 1: accidental add/commit? this file should not be in version control

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperListener.java
```diff
@@ -0,0 +1,36 @@
...
+import org.opencadc.posix.mapper.auth.DelegatingAPIKeyIdentityManager;
+
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+
+
+/**
+ * A bit of an ugly hack to reset the IdentityManager with a delegating one.
+ */
+public class PosixMapperListener implements ServletContextListener {
```
pdowler, line 15: This could be done in the InitAction.doInit() and doShutdown() which is intended for this kind of thing.

Yes, this listener mechanism exists, but the cadc-rest one allows you to take advantage of context or endpoint specific strings to make "more local" JNDI keys (if necessary).... it is not necessary in this case because the basic IdentityManager thing is based on java system prop config;  it's also the common place to look for service initialization so **I'd prefer it be in there**. Also, we can then control when this happens relative to other initialization and make use of other config in case that ever matters.

posix-mapper/src/main/webapp/WEB-INF/web.xml
```diff
@@ -5,8 +5,12 @@
         "http://java.sun.com/j2ee/dtds/web-app_2_3.dtd">
 
 <web-app>
-
     <display-name>posix-mapper</display-name>
+    
+    <listener>
+        <!-- One-time execution to preserve the Identity Manager. -->
```
pdowler, line 11: can be removed when listener code moved to InitAction

pdowler (changes requested):

posix-mapper/README.md
```diff
@@ -48,19 +75,18 @@ org.opencadc.posix.mapper.resourceID=ivo://{authority}/{name}
...
 org.opencadc.posix.mapper.schema=mapping
 
-# home dir root
-org.opencadc.posix.mapper.homeDirRoot=/storage/home
-
 # ID ranges to allow some customization where administration is necessary
 org.opencadc.posix.mapper.uid.start=10000
 org.opencadc.posix.mapper.gid.start=90000
+
+# At least one group that are allowed to query the API.  Use multiple org.opencadc.posix.mapper.group entries for
```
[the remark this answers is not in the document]
at88mph: The current implementation for this requires at least one `group` entry, or nobody will be allowed to use the system unless they pass an `api-key`.  I think we added this as a way to collect allowed users.

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperListener.java
at88mph: Sure.  We currently have two Servlets (one for UID and another for GID).  They both configure an `InitAction` for use now, which means this will be done twice, but it *must* be done only once.  Is it preferable to combine those two Servlets and have a single `GetAction`, or just create a dummy Servlet with an `InitAction` that only does this work?

posix-mapper/src/main/webapp/WEB-INF/web.xml
at88mph: Done.

posix-mapper/docker/config/posix-mapper/catalina.properties
at88mph: Yup.  Removed.

posix-mapper/build.gradle
```diff
@@ -26,6 +26,7 @@ war {
 
 dependencies {
     implementation 'org.hibernate.orm:hibernate-core:6.2.0.Final'
+    implementation 'org.opencadc:cadc-access-control-identity:[1.2.5,2.0)'
```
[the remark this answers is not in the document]
at88mph: Not required.

posix-mapper/README.md
```diff
@@ -23,6 +46,10 @@ for common system properties.
 - See <a href="https://github.com/opencadc/ac/tree/master/cadc-access-control-identity">cadc-access-control-identity</a> for CADC access-control system support.
 - See <a href="https://github.com/opencadc/ac/tree/master/cadc-gms">cadc-gms</a> for OIDC token support.
 
+| :point_down:  IdentityManager System Property                                                                                                                                                                                                                                                       |
+|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| The POSIX Mapper is a special case that allows API Keys for access.  As such, setting the `ca.nrc.cadc.auth.IdentityManager` System Property will be overwritten by an internal implementation.  This internal implementation will still delegate calls to the configured IdentityManager, however. |
```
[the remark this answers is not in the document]
at88mph: Done.

posix-mapper/README.md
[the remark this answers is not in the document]
at88mph: Is this accurate then?  It may be confusing for some when no Groups are set, I guess.  Helm will ensure at least one is set though.

pdowler:
the build fail is due to a new cadc-util OpenIDPrincipal ctor; the fix for cadc-access-control is in a PR but that also has incompatible changes to StandardIdentityManager and we havent' figured out versioning there. I'll figure something out right now

posix-mapper/README.md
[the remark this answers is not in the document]
pdowler: well, given the uncertainty of the whole thing, I guess we can leave it like this.

pdowler:
This ism sort of OBE (overcome by events) som it's going to sit as-is until we know form sure if we need apikeys or not.

PR #170, opened 2024-07-10
Added support for `OpenIdPrincipals` with different issuers

andamian:


cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -278,11 +288,12 @@ public Subject toSubject(Object owner) {
     @Override
     public Object toOwner(Subject subject) {
         // use NumericPrincipal aka OIDC sub for persistence
-        Set<NumericPrincipal> ps = subject.getPrincipals(NumericPrincipal.class);
+        Set<OpenIdPrincipal> ps = subject.getPrincipals(OpenIdPrincipal.class);
```
pdowler, line 291: comment above is now wrong; no comment needed since code is now 100% clear

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -244,18 +247,25 @@ public Subject augment(Subject subject) {
...
-            UUID uuid = null;
-            if (owner instanceof UUID) {
-                uuid = (UUID) owner;
-            } else if (owner instanceof String) {
-                String sub = (String) owner;
-                uuid = UUID.fromString(sub);
+            if (owner instanceof String) {
+                String[] openIDComponents = ((String)owner).split(OID_OWNER_DELIM);  // "issuer openID"
+                if (openIDComponents.length != 2) {
+                    throw new RuntimeException("unexpected owner format: " + owner.getClass().getName() + " value: " + owner);
```
pdowler, line 255: I wonder if we could retain backwards compatibility here by treating length==1 as a bare `sub` value, presumably from the single known oidcIssuer?? In  theory we could but it might just push the problem someplace else... we should discuss the next step(s) before deciding this and/or the version number (due to incompat).

pdowler (changes requested):

pdowler (approved):

PR #172, opened 2024-07-29
OIDC token validation using issuer public key (CADC-13465)

andamian:


brianmajor:
@pdowler - Should we review and merge this?

pdowler:
will review once the build passes

pdowler (changes requested):
Once the merge conflict is resolved, at least some of the extra code here should be moved into OIDCClient so it can benefit from the cached openid-configuration

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -296,86 +333,215 @@ public String toDisplayString(Subject subject) {
...
+                    } // else not a bearer token
+                } else {
+                    throw new NotAuthenticatedException(raw.getHeaderValue(), NotAuthenticatedException.AuthError.INVALID_REQUEST,
+                            "BUG: invalid authorization " + raw.getHeaderValue());
                 }
+            } // else other challenge
+        }
+    }
+
+    private void validateWithToken(Subject s, AuthorizationTokenPrincipal raw) {
```
pdowler, line 381: This method name could be simply `validateToken` since it does both validateWithPubKey and validateWithUserInfo... the `With` seems slightly misleading as it sounds like "using the token" which is really only the latter method

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -296,86 +333,215 @@ public String toDisplayString(Subject subject) {
...
+        String str = StringUtil.readFromInputStream(istream, "UTF-8");
+        JSONObject json = new JSONObject(str);
+        String sub = json.getString("sub");
+        String username = null;
+        if (json.has("preferred_username")) {
+            // jwt token
+            username = json.getString("preferred_username");
+        } else {
+            if (json.has("name")) {
+                // TODO not sure if this is correct but this is what the CADC userinfo has for the HttpPrincipal.
```
pdowler, line 470: `name` does seem really vague.... I could see `username` as an alternative. I wonder if this is done that way in CADC because that's what harbor expects? I guess OK for now

pdowler (changes requested):
If you look at the StandardIdentityManager intTest, you will see that if you create a file with a valid token (from the provider in the intTest/resources/cadc-registry.properties) you can test this... did you notice that and try it?

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -296,86 +333,215 @@ public String toDisplayString(Subject subject) {
...
-                        int max = Math.min(details.length(), 32);
-                        sb.append(details.subSequence(0, max));
-                        if (max < details.length()) {
-                            sb.append("...");
+        boolean validated = false;
+        for (AuthorizationTokenPrincipal raw : rawTokens) {
+            if (AuthenticationUtil.AUTHORIZATION_HEADER.equalsIgnoreCase(raw.getHeaderKey())) {
+                String[] tval = raw.getHeaderValue().split(" ");
+                if (tval.length == 2) {
+                    if ("Bearer".equalsIgnoreCase(tval[0])) {
```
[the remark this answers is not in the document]
andamian: Not being very familiar with tokens, I wasn't sure whether other types of tokens (not bearer) can be sent in the same request for other purposes (session tracking, credential delegation, etc). But it's probably better to start with minimum and open the faucet only when needed.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
andamian: That's what it is right now so it's the only way to keep it backwards compatible if that's a requirement.

pdowler (dismissed):

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -306,75 +327,201 @@ public String toDisplayString(Subject subject) {
...
+        String credentials = tval[1];
+        log.debug("challenge type: " + challengeType);
+        log.debug("credentials: " + credentials);
+
+        URI jwtIssuer = null;
+        try {
+            jwtIssuer = getJwtIssuer(credentials);
+            if (!jwtIssuer.normalize().equals(oidcClient.issuer.normalize())) {
+                log.debug("Token from untrusted issuer: " + jwtIssuer + " ignored");
+                jwtIssuer = null; // not the configured issuer, so don't use it
```
pdowler, line 376: by setting this to null, the validation falls through to trying userinfo endpoint on the trusted IDP... is that necessary if the this is a JWT and the issuer doesn't match? Should this throw NotAuthenticatedException right here?

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -306,75 +327,201 @@ public String toDisplayString(Subject subject) {
...
+
+    private List<Principal> validateWithPubKey(URI jwtIssuer, String challengeType,String credentials)
+            throws MalformedURLException, InvalidJwtException, MalformedClaimException {
+        VerificationKeyResolver httpsJwksKeyResolver = getHttpsJwksVerificationKeyResolver(jwtIssuer, challengeType);
+        JwtConsumer jwtConsumer = new JwtConsumerBuilder()
+                .setRequireExpirationTime()
+                .setExpectedIssuers(true, jwtIssuer.toString())
+                .setVerificationKeyResolver(httpsJwksKeyResolver)
+                .setJwsAlgorithmConstraints(new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.PERMIT, // whitelist
+                        AlgorithmIdentifiers.RSA_USING_SHA256)) // RS256 only
```
pdowler, line 419: Why just RS256? the config of SRC IAM says it supports a list of encryption and signing algorithms? Can't you tell from the JWT which was used? If not, does the coe have to just try all the advertised algs? ugh

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -306,75 +327,201 @@ public String toDisplayString(Subject subject) {
...
+
+        InputStream istream = get.getInputStream();
+        String str = StringUtil.readFromInputStream(istream, "UTF-8");
+        JSONObject json = new JSONObject(str);
+        String sub = json.getString("sub");
+        String username = null;
+        if (json.has("preferred_username")) {
+            // jwt token
+            username = json.getString("preferred_username");
+        } else {
```
pdowler, line 453: `else if {` ??

and since we ran into the before (well, Rubin did), maybe a final `else {` with a debug statement that there is no username would be a good addition here.

also note that on line 432 you only check for `(jwtClaims.getClaimValueAsString("preferred_username)`... is that more strictly defined?

pdowler (changes requested):

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
andamian: Left the `name` option strictly for the CADC case.

pdowler (approved):

PR #173, opened 2024-10-01
Userinfo lookup

at88mph:
Added client class to query configured OpenID Connect provider to be built upon for future uses.

### DIFF
--- a/cadc-gms/src/main/java/org/opencadc/auth/OIDCClient.java
+++ b/cadc-gms/src/main/java/org/opencadc/auth/OIDCClient.java
@@ -83,7 +83,7 @@ import org.json.JSONObject;
 
 
 /**
- * Client to the currently configured OpenID Connect provider.
+ * Client to the currently configured OpenID Connect provider.  This Client class is intended to have a very short lifespan.
  * TODO: Generic enough to be more central?
  *
  * @see LocalAuthority for configuration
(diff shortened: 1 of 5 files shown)


at88mph:
CADC-13755

Testing this now with the Science Platform

cadc-gms/src/main/java/org/opencadc/auth/OIDCClient.java
```diff
@@ -0,0 +1,159 @@
...
+
+
+/**
+ * Client to the currently configured OpenID Connect provider.
+ * TODO: Generic enough to be more central?
+ *
+ * @see LocalAuthority for configuration
+ * @see ca.nrc.cadc.reg.client.RegistryClient for configuration
+ */
+class OIDCClient {
```
pdowler, line 92: For a class named "client" I would rather see a constructor that took the issuer URI as a ctor arg and leave the config lookup in the hands of the calling code. Then non-static methods like a client normally provides.

That would be more future-proof and reusable and we'd be able to merge the PR 172 use cases in and end up with less code and fewer calls to get the well known config. Also makes the `TODO: Cache this?` question below have an obvious answer of yes.

With these changes, my review of the other PR will be easy: "move these methods into OIDClient" :-)

pdowler (changes requested):

cadc-gms/src/main/java/org/opencadc/auth/OIDCClient.java
at88mph: OK.  This class gets instantiated by the `StandardIdentityManager`, which, I think, means its lifespan is very short.  I've put a dumb cache in place with that assumption in mind.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -132,19 +129,15 @@ public class StandardIdentityManager implements IdentityManager {
...
-            URL u = oidcIssuer.toURL();
-            oidcDomains.add(u.getHost());
+        oidcClient = new OIDCClient(loc.getServiceURI(key));
 
-            // add known and assume trusted A&A services
-            String host = getProviderHostname(loc, Standards.GMS_SEARCH_10);
-            if (host != null) {
-                oidcDomains.add(host);
-            }
+        URL u = oidcClient.getIssuerURL();
```
pdowler, line 134: This is where the call to get well know config happens and it occurs for every new IM instance. **That is going to be costly** - previous impl did zero calls on instantiation (well, reg calls but they are cached on local machine for ~10 min) and then probably one userinfo call and optionally one posix-mapper call later on.

For reasons, the AuthenticationUtil.getIdentityManager() call that instantiates an IM can happen from multiple places within a request and each of those will suffer this overhead.
---
I'm wondering if OIDCClient can re-use CachingFile from cadc-registry to cache the well-known config in the filesystem... something like $TMP/{user}/cadc-gms/{issuer hostname}/{path from url}

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -388,15 +381,7 @@ private void validateOidcAccessToken(Subject s) {
     
     private URL getUserEndpoint() {
```
pdowler, line 382: just remove this method entirely and replace it's use with the oidcClient call

pdowler (changes requested):

cadc-gms/src/main/java/org/opencadc/auth/OIDCClient.java
```diff
@@ -0,0 +1,165 @@
...
+                final Reader inputReader = new BufferedReader(new InputStreamReader(inputStream));
+                final char[] buffer = new char[8192];
+                int charsRead;
+                while ((charsRead = inputReader.read(buffer)) >= 0) {
+                    writer.write(buffer, 0, charsRead);
+                }
+                writer.flush();
+            });
+
+            httpGet.run();
```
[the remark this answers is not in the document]
at88mph: Now relies on the `CachineFile` `GET` request.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
at88mph: Done.

pdowler (dismissed):

pdowler (approved):

PR #174, opened 2024-12-23
fix: fix memory leak with a single client instance

at88mph:
some cleanup
tested with a visualvm profiler to show heap space being garbage collected
https://visualvm.github.io/

### DIFF
--- a/posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperAction.java
+++ b/posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperAction.java
@@ -74,11 +74,7 @@ import ca.nrc.cadc.auth.NotAuthenticatedException;
 import ca.nrc.cadc.rest.InlineContentHandler;
 import ca.nrc.cadc.rest.RestAction;
 import ca.nrc.cadc.util.MultiValuedProperties;
-import org.opencadc.posix.mapper.Group;
-import org.opencadc.posix.mapper.PosixClient;
-import org.opencadc.posix.mapper.Postgres;
-import org.opencadc.posix.mapper.PostgresPosixClient;
-import org.opencadc.posix.mapper.User;
+import org.opencadc.posix.mapper.*;
 import org.opencadc.posix.mapper.web.group.AsciiGroupWriter;
 import org.opencadc.posix.mapper.web.group.GroupWriter;
 import org.opencadc.posix.mapper.web.group.TSVGroupWriter;
@@ -94,18 +90,22 @@ import java.io.Writer;
 
 public abstract class PosixMapperAction extends RestAction {
 
-    protected PosixClient posixClient;
     protected static final MultiValuedProperties POSIX_CONFIGURATION = PosixInitAction.getConfig();
     protected static final String TSV_CONTENT_TYPE = "text/tab-separated-values";
 
+    // As the Postgres class uses a Factory pattern, a static client is created to prevent multiple factory instances
+    // being created that was causing a memory leak.
+    // jenkinsd (2024.12.23)
+    //
+    protected static final PosixClient POSIX_CLIENT =
+            new PostgresPosixClient(Postgres.instance(PosixMapperAction.POSIX_CONFIGURATION
+                                                              .getFirstPropertyValue(PosixInitAction.SCHEMA_KEY))
+                                            .entityClass(User.class, Group.class)
+                                            .build());
 
     protected PosixMapperAction() {
-        final Postgres postgres = Postgres.instance(PosixMapperAction.POSIX_CONFIGURATION
-                                                            .getFirstPropertyValue(PosixInitAction.SCHEMA_KEY))
-                                          .entityClass(User.class, Group.class)
-                                          .build();
-        this.posixClient = new PostgresPosixClient(postgres);
     }
+
 
     @Override
     public void initAction() throws Exception {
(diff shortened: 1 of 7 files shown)


at88mph:
@pdowler Can you review or delegate it?

cadc-permissions-client/build.gradle
```diff
@@ -14,17 +14,18 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '0.3.1'
+version = '0.3.2'
```
pdowler, line 17: Please revert this as it is unrelated to the rest of the PR.

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperAction.java
```diff
@@ -74,11 +74,7 @@
 import ca.nrc.cadc.rest.InlineContentHandler;
 import ca.nrc.cadc.rest.RestAction;
 import ca.nrc.cadc.util.MultiValuedProperties;
-import org.opencadc.posix.mapper.Group;
-import org.opencadc.posix.mapper.PosixClient;
-import org.opencadc.posix.mapper.Postgres;
-import org.opencadc.posix.mapper.PostgresPosixClient;
-import org.opencadc.posix.mapper.User;
+import org.opencadc.posix.mapper.*;
```
pdowler, line 77: We do not use wildcard imports

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/group/GetAction.java
```diff
@@ -81,8 +81,8 @@ public class GetAction extends PosixMapperAction {
     @Override
     public void doAction() throws Exception {
         final GroupWriter groupWriter = getGroupWriter();
-        this.posixClient.writeGroups(groupWriter, groupParameters().toArray(new GroupURI[0]),
-                                     gidParameters().toArray(new Integer[0]));
+        GetAction.POSIX_CLIENT.writeGroups(groupWriter, groupParameters().toArray(new GroupURI[0]),
```
pdowler, line 84: Reference to a static should use the class that declared it, not a subclass.

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/user/GetAction.java
```diff
@@ -81,8 +81,8 @@ public class GetAction extends PosixMapperAction {
     @Override
     public void doAction() throws Exception {
         final UserWriter userWriter = getUserWriter();
-        posixClient.writeUsers(userWriter, usernameParameters().toArray(new String[0]),
-                               uidParameters().toArray(new Integer[0]));
+        GetAction.POSIX_CLIENT.writeUsers(userWriter, usernameParameters().toArray(new String[0]),
```
pdowler, line 84: As above, unclear use of static POSIX_CLIENT

pdowler (changes requested):

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperAction.java
```diff
@@ -94,19 +94,23 @@
...
 
-    protected PosixClient posixClient;
     protected static final MultiValuedProperties POSIX_CONFIGURATION = PosixInitAction.getConfig();
     protected static final String TSV_CONTENT_TYPE = "text/tab-separated-values";
 
+    // As the Postgres class uses a Factory pattern, a static client is created to prevent multiple factory instances
+    // being created that was causing a memory leak.
+    // jenkinsd (2024.12.23)
+    //
+    protected static final PosixClient POSIX_CLIENT =
```
pdowler, line 104: Are you certain that PostgresPosixClient (and it's member Postgres) are 100% thread-safe? That's the risk of declaring a static (or storing a shared instance in JNDI) - it gets used by mutliple request threads and has to be thread-safe.

cadc-permissions-client/build.gradle
pdowler: this whole file change still needs to be reverted

pdowler (changes requested):

cadc-permissions-client/build.gradle
at88mph: That's done.  Did you see my earlier comment that it fixes the CI build though?  It's not related, true, but is it fixed elsewhere?

posix-mapper/src/main/java/org/opencadc/posix/mapper/web/PosixMapperAction.java
at88mph: True, and I'm not 100%.  I was able to use four clients hitting the service with three uid and gid entries without fail in a continuous loop.  The problem it solves is having a single Hibernate `SessionFactory` instance, which is thread safe:
https://docs.jboss.org/hibernate/core/3.5/api/org/hibernate/SessionFactory.html

I'm not sure if there is a better way to do this without Spring injecting it, or maybe having some JNDI lookup for it?

pdowler: As long as the trail of stateful objects are either immutable or themselves threadsafe, then it should be ok. 

Using JNDI instead of static doesn't make it more threadsafe, it just effects the scope of the instance:

static: whole JVM
JDNI: can be scoped to a single app or a single endpoint by how the JNDI key is specified

That doesn't matter here because this is service code, not library code and only deployed as a container with one app in the jvm.

PR #179, opened 2025-06-12
Automatic user account creation with OpenID authentication (CADC-14107)

andamian:


cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
```diff
@@ -133,15 +137,20 @@ public Set<URI> getSecurityMethods() {
 
     @Override
     public Subject validate(Subject subject) throws NotAuthenticatedException {
-        return TokenValidator.validateTokens(subject);
+        Subject sub = TokenValidator.validateTokens(subject);
+        if (!sub.getPrincipals(AuthorizationTokenPrincipal.class).isEmpty()) {
+            StandardIdentityManager sim = new StandardIdentityManager();
```
pdowler, line 142: This can fail with a runtime exception if there is no configured OpenID provider; I feel like that is a sane config so that exception probably needs to be caught and logged (debug).

cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
```diff
@@ -229,44 +242,58 @@ public Object toOwner(Subject subject) {
             if (principal instanceof X500Principal) {
                 x500Principal = (X500Principal) principal;
             }
+            if (principal instanceof OpenIdPrincipal) {
+                openIdPrincipal = (OpenIdPrincipal) principal;
+            }
         }
 
-        if (x500Principal == null) {
+        NumericPrincipal numericPrincipal;
```
pdowler, line 250: In principal, a user could authenticate with an unknown OpenID *and* client cert... createAuthUser could make sure both are included in the new account, eg change to `createAuthUser(subject)`. Currently the code is chosing one type over the other.

cadc-access-control-identity/src/test/java/ca/nrc/cadc/ac/ACIdentityManagerTest.java
```diff
@@ -106,6 +107,7 @@ public void testCertOnlyUser() {
             Subject subject = new Subject();
             subject.getPrincipals().add(new X500Principal(dn));
             subject.getPublicCredentials().add(AuthMethod.CERT);
+            System.setProperty(PropertiesReader.CONFIG_DIR_SYSTEM_PROPERTY, "build/resources/test/config");
```
pdowler, line 110: if you set a system prop, you really need a try/finally so you can unset it properly after this test; otherwise it can effect other tests in odd ways (because of how junit works, gradle daemon mode, etc).

pdowler (changes requested):
just reviewed cadc-access-control-identity so far

cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
pdowler: Not quite: at this point both the OpenID and X509 identifies are _new_ so it is not linking; it is someone providing two identities in one go.

If someone authenticated with both a token and cert and the OpenID was _new_ but the cert was not we would ignore the OpenID (there would be a NumericPrincipal). Likewise, if the cert was _new_ but the OpenID was not we would ignore the cert. Is that OK or bad? But _ignore_ here means it might still be in the subject, just not linked/interchangeable with other identifies in the same subject...

andamian: I can see problems down the road if we allow bot cert and token. The second authentication might be used downstream and could trigger a different automatic account creation. Maybe I should check for it everywhere and fail it until we figure out a way to deal with this case.

pdowler: I think accepting it fully or failing are both better than silently chosing one ident and ignoring the other, even though this is a very obscure usage. I'd be OK with fail for now.

andamian: Done

cadc-access-control-identity/src/main/java/ca/nrc/cadc/ac/ACIdentityManager.java
```diff
@@ -133,15 +137,25 @@ public Set<URI> getSecurityMethods() {
 
     @Override
     public Subject validate(Subject subject) throws NotAuthenticatedException {
-        return TokenValidator.validateTokens(subject);
+        Subject sub = TokenValidator.validateTokens(subject);
+        if (!sub.getPrincipals(AuthorizationTokenPrincipal.class).isEmpty()) {
+            try {
+                StandardIdentityManager sim = new StandardIdentityManager();
+                return sim.validate(sub); // validates and resolves JWT tokens
+            } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
andamian: Removed `catch`

cadc-access-control-server/src/main/java/ca/nrc/cadc/ac/server/web/users/CreateUserAction.java
```diff
@@ -83,12 +89,51 @@ public class CreateUserAction extends AbstractUserAction {
...
+            log.debug("Can't self-create CADC account: no subject or principals");
+            return false; // no subject or principals
+        }
+        Set<OpenIdPrincipal> validPrinc = sub.getPrincipals(OpenIdPrincipal.class);
+
+        if (validPrinc.size() > 1) {
+            log.debug("Can't self-create user account. Multiple OpenID Identities: " + validPrinc);
+            return false; // more than one OpenID principal is not allowed
+        }
+        if (validPrinc.size() == 1) {
```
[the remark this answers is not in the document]
andamian: I've now drastically reduced the self create abilities - only users with OpenID and X500 principals can self create users provided that the unique identity of the subject matches the identity of the requested user. I feel like we need to be restrictive given that there is no account curation through this route. If there are valid cases where this proof to be too restrictive then we can look again.

PR #181, opened 2025-07-17
fix: skip remote validation of custom auth tokens

at88mph:
When using AuthorizationTokens that are not `Bearer` or `Basic`, they cannot be validated remotely, so simply pass them into the Subject.

### DIFF
--- a/cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
+++ b/cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
@@ -167,7 +167,7 @@ public class StandardIdentityManager implements IdentityManager {
 
     @Override
     public Subject validate(Subject subject) throws NotAuthenticatedException {
-        validateAuthorizationTokens(subject);
+        validateOidcAccessToken(subject);
         return subject;
     }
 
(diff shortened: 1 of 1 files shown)


cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -335,43 +334,53 @@ private void validateOidcAccessToken(Subject s) {
...
+                            if (max < details.length()) {
+                                sb.append("...");
+                            }
+                            throw new NotAuthenticatedException(challengeType, NotAuthenticatedException.AuthError.INVALID_TOKEN, sb.toString());
+                        } catch (Exception ex) {
+                            throw new NotAuthenticatedException(challengeType, NotAuthenticatedException.AuthError.INVALID_TOKEN, ex.getMessage(), ex);
+                        }
+                    } else {
+                        // Otherwise it's a custom token, so we just create an AuthorizationToken and let whatever
+                        // application is looking for it handle it.
```
pdowler, line 379: No. I thought it was clear that this IM impl should just leave the other ATP alone. If some other code does not deal with it, then authentication is rejected: that is the intended behaviour.

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
```diff
@@ -307,7 +306,7 @@ public String toDisplayString(Subject subject) {
         return null;
     }
 
-    private void validateOidcAccessToken(Subject s) {
```
pdowler, line 310: This method rename is unnecessary/misleading so should be reverted: the original name is correct.

pdowler (changes requested):

cadc-gms/src/main/java/org/opencadc/auth/StandardIdentityManager.java
at88mph: Will fix.

pdowler (approved):

PR #185, opened 2025-08-06
Add access module and integrate into CI build

WenbinWL:
# CADC Access Module

This module provides authentication and access control components for CADC web-based services. It was migrated from the legacy internal repository (`gimli:/srv/cadc/git/wwebui/access`) into the GitHub monorepo [`opencadc/ac.git`](https://github.com/opencadc/ac).

---

##  Module Overview

- Java WAR application for access control
- Originally written for deployment in CADC web environments
- Uses `cadc-util`, `cadc-registry`, and other CADC libraries
- Legacy tests built using EasyMock

---

### Requirements

- Java 11
- Gradle (use the provided `gradlew`)

### Build the WAR file

```bash
./gradlew :access:build

### CODE REVIEW

File: access/access/README.md [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+
+For production, generate keys using:
+```bash
+# Generate 2048-bit RSA key pair
+openssl genrsa -out RsaSignaturePriv.key 2048
+openssl rsa -in RsaSignaturePriv.key -pubout -out RsaSignaturePub.key
+```
+
+## Deployment
+
```
pdowler, line 63: the baldur/README has the standard recommended commands to build, check, and run; that normally means the current dir is the module and not the repo base dir. 

although we include gradle wrapper in the repo, devs are free to use it or gradle (installed in some other way) as they see fit, or gradle within an IDE

access/src/test/java/ca/nrc/cadc/accesscontrol/GenerateTestKey.java
```diff
@@ -0,0 +1,46 @@
...
+import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.util.Base64;
+
+/**
+ * Generate test RSA key pair
+ */
+public class GenerateTestKey {
+
```
pdowler, line 14: This whole class is not really necessary. I think you could have added a single line to the test setup (like a static setup method in the abstract test class) with a call to the code in cadc-util that I mentioned. There's no need to replicate all that stuff here. There is no need to implement a `main()` and it's certainly less desireable than having it run automatically at the start of test execution (and put keys in the `build` dir rather)

pdowler (changes requested):
It looks like you accidentally committed `.gradle` (cache); that needs to be removed
Likewise, please don't commit .DS_Store files. I also see a class file in the src/test tree...

I'm not sure why there is an extra `access` directory. Should just be:
```
ac.git/access/README.md
ac.git/access/build.gradle
ac.git/access/Dockerfile
ac.git/access/VERSION
ac.git/access/src
```

In the CI, you just need to add this module to the existing gradle.yml in the same style as other build. Since this is a web service (not a library), this should suffice:
```
../gradlew --info clean build
```
eg javadoc only needed for libs and checkstyleMain is technical debt for another day.

pdowler:
replaced by PR #187

PR #202, opened 2026-01-21
New GMS service specification

andamian:


gms/src/main/webapp/openapi/openapi.yaml
```diff
@@ -0,0 +1,605 @@
...
+              $ref: '#/components/schemas/Group'
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Group'
+      responses:
+        '200':
+          description: Group created successfully
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '401':
```
pdowler, line 128: several common responses are defined in vosi and can be reused with a $ref (401, 403, 404 at least)

I do not think 404 for member not found is really correct and we should chose a different response code for that.

gms/src/main/webapp/openapi/openapi.yaml
```diff
@@ -0,0 +1,605 @@
...
+        '503':
+          $ref: '#/components/responses/ServiceUnavailable'
+
+    put:
+      summary: Create group
+      description: Create a new group according to the group XML or JSON document in the HTTP PUT
+      operationId: createGroup
+      tags:
+        - Group Management
+      security:
```
pdowler, line 110: I have generally not been including security spec in the APIs. I think it restricts but does not add anything useful.

gms/src/main/webapp/openapi/openapi.yaml
```diff
@@ -0,0 +1,605 @@
...
+                  group1
+                  group2
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+        '503':
+          $ref: '#/components/responses/ServiceUnavailable'
+
+    put:
```
pdowler, line 104: In good RESTful API, put means the caller decides on the name of the resource to create, so this `put` should be to the path `/groups/{name}`.

gms/src/main/webapp/openapi/openapi.yaml
```diff
@@ -0,0 +1,605 @@
...
+      tags:
+        - Group Management
+      security:
+        - clientCertificate: []
+        - bearerAuth: []
+      responses:
+        '200':
+          description: Group member removed successfully
+
+  /search:
```
pdowler, line 379: this should be a $ref to `openapi/gms/search.yaml`

pdowler (commented):
minor organisation: in `youcat` I put openapi.yaml and openapi (dir) into webapps as siblings... 

if you look at the TAP openapi.yaml in my fork:
https://github.com/pdowler/ivoa-TAP/blob/main/openapi.yaml

you will see the advantage of splitting each path component out into a separate file -- the big picture path structure becomes immediately apparent.

---

I also stopped at line 268
```
/groups/{groupID}/user-members/{userID}:
```
because I think we should review this aspect of the API (adding and removing members). I take it that this describes what we have implemented already? This reminds me of UWS where there are a bunch of child resources under the job and it isn't my favourite style. Still, if you used $ref for every path element this approach would be more clear and could still be the right choice. It's just hard to tell.

PR #212, opened 2026-04-20
simple AAI data model

pdowler:
to support posix-mapper and full user and group concepts


PM/openapi/posix/posix-groupmap.yaml
```diff
@@ -0,0 +1,105 @@
...
+      $ref: ../vosi/vosi-std-responses.yaml#/permission-denied
+    '404':
+      $ref: ../vosi/vosi-std-responses.yaml#/not-found
+post:
+  operationId: create-groupmap-entry
+  tags:
+    - POSIX Groups
+  summary: create a local posix info for a group
+  parameters:
+    - name: user
```
andamian: Name `user`? And maybe more details? For example it should be the full URI of a group in one of the configured GMS services.

pdowler: fixing

PR #213, opened 2026-04-22
Re-worked OpenAPI spec (CADC 15042)

andamian:
Renamed branch


gms/src/main/webapp/openapi/gms/gms-search.yaml
```diff
@@ -0,0 +1,47 @@
...
+    contain only those groups for which membership is confirmed.
+
+    If no `group` parameter is supplied, the response SHALL contain all
+    groups of which the identity is a member.
+
+    The response SHALL be plain text containing zero or more lines,
+    each line containing a simple group name.
+  operationId: getMembership
+  parameters:
+    - $ref: ../components.yaml#/components/parameters/group
```
pdowler: if this is imported (copied) from  ivoa-std/GMS then it can't refer to a local resource... I'd prefer a copy for now and we'll have to figure out how to sanely extend an API. Clearly one can add endpoints (because new paths go in base openapi.yaml).

I think adding custom params is either some kind of subclassing (don't know if that is a thing) or a fork we have to keep in sync with upstream... which is hard.

gms/src/main/webapp/openapi/paths/groups-group-members.yaml
```diff
@@ -0,0 +1,42 @@
...
+      description: Group member added successfully
+    '400':
+      $ref: ../vosi/vosi-std-responses.yaml#/bad-request
+    '401':
+      $ref: ../vosi/vosi-std-responses.yaml#/not-authenticated
+    '403':
+      $ref: ../vosi/vosi-std-responses.yaml#/permission-denied
+    '404':
+      description: One or both groups could not be found
+    '500':
```
pdowler: I have thus far not been including these in openapi specs, but it's probably a good idea... they could probably go into vosi-responses.yaml for max reusability

gms/src/main/webapp/openapi/paths/groups-groupid.yaml
```diff
@@ -0,0 +1,90 @@
+get:
```
pdowler: this file could just be named group.yaml (ops on a single group)... having "groups" in front ties it to the path and that's not necessary or flexible

same with other two files prefixed by "groups-"

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
+# Shared components for the GMS OpenAPI description (referenced from path fragments and the root document).
+components:
+  parameters:
```
pdowler, line 3: parameter definition style:

I found one of:
```
- name: groupID
  in: path
...
```
or (eg)
```
- $ref: {defined elsewhere}
```
to be the most compact. See VOSI.git for examples of params defined in separate files.

Having 
```
groupID:
  name: groupID
...
```
is either redundant or confusing when the two values differ.

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
...
+    groupID:
+      name: groupID
```
pdowler: I would probably use `groupName` so no one makes assumptions that are wrong. When I see `groupID` I think it is going to be the URI defined by GMS.

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
...
+      in: path
+      required: true
+      description: The name of the group
+      schema:
+        type: string
+
+    userID:
```
pdowler, line 13: This one is tricky because we can't say much about the values; need to think/discuss ...

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
...
+      name: userID
+      in: path
+      required: true
+      description: The unique user identifier (CANFAR HTTP username)
+      schema:
+        type: string
+
+    group:
```
pdowler: I think this param does not belong here

gms/src/main/webapp/openapi.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - **CC** - Client certificates over HTTPS
+    - **JWT** - JWT Bearer token over HTTPS
+    - **AN** - Anonymous over HTTP
+
+    **User Identity:**
+    Users are identified by their unique userID in the Identity Provider (CANFAR HTTP username).
+  version: 1.0.0
+  contact:
+    url: https://www.canfar.net
+tags:
```
pdowler, line 26: the order of global tags controls then order things are rendered by swagger; I would need to see it to think about what order is best

gms/src/main/webapp/openapi.yaml
```diff
@@ -0,0 +1,59 @@
...
+    description: Operations for searching groups by user role (IVOA GMS 1.0)
+  - name: Group Membership
+    description: GMS membership and identity-related queries
+  - name: VOSI Capabilities
+    description: VOSI capability and probe operations
+  - name: Service availability
+    description: Service availability
+  - name: POSIX Groups
+    description: POSIX gid / group name search
+servers:
```
pdowler, line 39: Should be `-url: /gms` without a server name and fix or remove description (in youcat I don't have one).

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
...
+        - ivo://example.org/gms?engineering
+      schema:
+        type: array
+        items:
+          type: string
+          example: ivo://example.org/gms?engineering
+      style: form
+
+  responses:
+    BadRequest:
```
pdowler: You don't need to and probabyl should not define things that come from external standards (eg Unauthorized and Forbidden which vosi defines)

In general I found that each endpoint needs to provide a description of 400 aka BadRequest so a generic one isn't really useful. 

Probably 500 and 503 should be in VOSI, so I don't think any of these response descriptions are useful.

gms/src/main/webapp/openapi/components.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+    InternalServerError:
+      description: The server encountered an unrecoverable error
+
+    ServiceUnavailable:
+      description: The server is too busy to perform the operation
+
+  schemas:
+    Group:
+      type: object
```
pdowler, line 36: I'm not a fan of trying to replicate the data model in openapi - it's quickly painful and hard to maintain.
TBD

gms/src/main/webapp/openapi/cadcvosi/availability.yaml
```diff
@@ -0,0 +1,31 @@
+get:
+  operationId: read-availability
```
pdowler, line 2: VOSI-availability was dropped from WD-VOSI-1.2 and no longer requried by ivoa services (details tbd).

We will fork and evolve availability to suit our own (OpenCADC) purposes, so it should be an imported component. Definitive source: TBD so this is just a note to think about it.

pdowler (changes requested):

gms/src/main/webapp/openapi/paths/groups-groupid.yaml
andamian: Replaced with single `group.yaml` file.

gms/src/main/webapp/openapi/components.yaml
andamian: That works in service parameters but not in components. In components the first occurrence is a key and the second one is the name. Although appears to be a repetition, that is not necessarily so as we could use a different value such as `groupIDKey` for example. Would that make the intention more obvious?

gms/src/main/webapp/openapi/components.yaml
andamian: I've added the `idType` parameter to conform to the current implementation. That could push the decision for later.

gms/src/main/webapp/openapi/components.yaml
andamian: Replaced with examples

gms/src/main/webapp/openapi/cadcvosi/availability.yaml
andamian: Moved to a different "source".

gms/src/main/webapp/openapi.yaml
andamian: Swagger is in the next story. We could fix it then.

PR #215, opened 2026-04-28
feat: api to authorise plugin and route permissions

at88mph:
# Changes
- New Permissions API Client
- Check returned JSON document for proper `is_authorised` key


cadc-permissions-client/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
     api 'org.opencadc:cadc-permissions:[0.2,)'
     implementation 'org.opencadc:cadc-registry:[1.0,2.0)'
     implementation 'org.opencadc:cadc-util:[1.6,2.0)'
+    implementation 'org.json:json:20251224'
```
pdowler: cadc-util includes an api dependency on this lib, which gives us control of external lib upgrades in a common place... so this dependency can/should be removed.

cadc-permissions-client/src/test/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClientTest.java
```diff
@@ -0,0 +1,101 @@
...
+ *  (c) 2026
+ *  Government of Canada                 Gouvernement du Canada
+ *  National Research Council            Conseil national de recherches
+ *
+ ************************************************************************
+ */
+
+package org.opencadc.permissions.client.srcnet;
+
+import com.sun.net.httpserver.HttpServer;
```
pdowler: This import of `com.sun` sets off alarm bells.

In general I am not a big fan of mocked unit tests and would prefer this to be an intTest that calls a real PAPI service (despite the currently ad-hoc and manual way that has to be done to authenticate properly). See the intTests in cadc-gms for example... we can discuss a common approach to such client tests to find something sane.

finally, I did allow some mock-http-server style tests in cadc-registry (with a caveat that if they are too hard to maintain I'll want to remove them, because that's my general experience with mocking)... I was convinced it was worth it in that case because the changes and tests are all about handling failures. In that case, a specific mock-server library was used rather than internal jvm code.

---

So, how about an intTest instead and we figure out how to handle credentials??

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private final URL baseURL;
+
+    /**
+     * Resolve the Permissions API base URL from the registry.
+     *
+     * @param baseServiceURL  Base URL to the PermissionsAPI service
+     */
+    public PermissionsAPIClient(final URL baseServiceURL) {
+        Objects.requireNonNull(baseServiceURL, "Base service URL cannot be null");
```
pdowler: I haven't used this in the past - what exception does this throw? If this happens in server side code, is it going to be a 500 (probably correct) directly or does the service have to catch this and improve the error message??

maybe document with `@throws` even though it is some kind of runtime exception

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
```diff
@@ -0,0 +1,169 @@
...
+    public AuthorisationResult authorisePlugin(final String serviceName, final String token,
+                                               final JSONObject requestBody, final String version) throws IOException {
+        assertArg(serviceName, "serviceName");
+        assertArg(token, "token");
+        final JSONObject body = requestBody != null ? requestBody : new JSONObject();
+        final URL url = buildPluginURL(serviceName, token, version);
+        final JSONObject json = PermissionsAPIClient.postBody(url, body);
+        try {
+            return AuthorisationResult.parse(json);
+        } catch (JSONException e) {
```
pdowler, line 159: In each of these authorise methods, a MalformedURLException is a bug, so they should catch that and
```
throw new RuntimeException("BUG: failed to create valid Permissions API request", ex);
```

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
```diff
@@ -0,0 +1,169 @@
...
+            throw new IOException("invalid JSON response", e);
+        }
+    }
+
+    private URL buildPluginURL(final String serviceName, final String token, final String version)
+            throws MalformedURLException {
+        String q = PermissionsAPIClient.queryTokenVersion(token, version);
+        String path = "v1/authorise/plugin/" + PermissionsAPIClient.encodePathSegment(serviceName);
```
pdowler: Q. Does Permissions API really require url-encoding path segments? It is quite a risky thing to specify...

pdowler (changes requested):
Main issue: prefer intTest vs real service over unit test with mock

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
at88mph: This actually throws a `NullPointerException`, but the caller gets to pick the message now.

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
at88mph: It was more for sanitization reasons.  I don't see any current service names with spaces in them, and the examples in the OpenAPI seem sane enough, so it may not be needed.  What is the risk?  A double-encoded `%` or something similar?

pdowler: Frameworks (eg java Servlet for us) decode parameters automatically, so no one ever writes applicationm code to do that. They do not decode path segments, so if special chars are ever used there it is up to applications to do it.

The two really risky characters are % and +: you have to encode and decode them the same number of times or the content is changed. And we know astronomers like to put + into names of things... 

A client can only encode path segment if the server API explicitly says to do so (because it is going to decode exactly once) - because other chars that are actually safe will get encoded even if they don't need to for path usage. 

For example, say PAPI includes a URI like `foo:bar` and they do not decode because that is a valid path segment in a URL. The encoded value will be `foo%3Abar` and not match.

cadc-permissions-client/src/main/java/org/opencadc/permissions/client/srcnet/PermissionsAPIClient.java
pdowler: An NPE is a RuntimeException so will result in a 500 in a service response. The problem is likely missing config so the service is the only place that can do any better with error messaging. Good enough as-is.

pdowler (dismissed):

at88mph:
Some additional changes:
- Authorize a token exchange for the requested service
- Use the `audience` from the authorization to exchange the token with the auth API
- Use the exchanged token for the route/plugin authorization

pdowler (approved):

